PR #27, opened 2015-06-21
Fix image scaling to 100% of screen width

tuomas2:
Here is a simple fix for image scaling bug https://github.com/mjdenham/and-bible/issues/24 . Scales image to 100% of the screen width. <del>I did not find any CSS support in AndBible so I added CSS style straight into  tag.</del>

### DIFF
--- a/AndBible/assets/web/style.css
+++ b/AndBible/assets/web/style.css
@@ -26,7 +26,7 @@ span[id="0"].verseNo + h1:after {
 }
 
 img.sword {
-	width: 100%;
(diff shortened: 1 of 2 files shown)


tuomas2:
Now I found also CSS, so I improved this PR such that style can be defined in CSS.

mjdenham:
I think I like this too.  It works well in portrait or smaller screens, but if viewing a small image in landscape it can expand too much so you only see a small part of the top of the image.  I will include it in a test build.  I think users will like it.

PR #28, opened 2015-06-22
WIP Show inline reference content always with genbooks and commentaries

tuomas2:
Fixes https://github.com/mjdenham/and-bible/issues/25

From what I understood from our discussion https://groups.google.com/d/msg/and-bible/cSZmdzDU1u0/L9lEaHaCnoQJ, current solution to not show reference content was to fix bible crossreferences in certain bible modules. My solution is essentially the same as in Xulsword, i.e. " But `<reference>` tags which are inline in regular text (such as commentaries) show the reference text as a link."

### DIFF
--- a/and-bible/app/src/main/java/net/bible/service/format/osistohtml/OsisToHtmlParameters.java
+++ b/and-bible/app/src/main/java/net/bible/service/format/osistohtml/OsisToHtmlParameters.java
@@ -29,7 +29,9 @@ public class OsisToHtmlParameters {
 	private BookmarkStyle defaultBookmarkStyle = BookmarkStyle.YELLOW_STAR;
     private boolean isShowNotes = false;
     private boolean isAutoWrapUnwrappedRefsInNote = false;
-    // used as a basis if a reference has only chapter and no book
+    private boolean isShowReferenceContent = true;
+
+	// used as a basis if a reference has only chapter and no book
     private Verse basisRef;
     private Versification documentVersification;
     private String font;
@@ -118,6 +120,12 @@ public class OsisToHtmlParameters {
 	public void setAutoWrapUnwrappedRefsInNote(boolean isAutoWrapUnwrappedRefsInNote) {
 		this.isAutoWrapUnwrappedRefsInNote = isAutoWrapUnwrappedRefsInNote;
 	}
+    public boolean isShowReferenceContent() {
+		return isShowReferenceContent;
+	}
+	public void setShowReferenceContent(boolean isShowReferenceContent) {
+		this.isShowReferenceContent = isShowReferenceContent;
+	}	
 	public boolean isShowStrongs() {
 		return isShowStrongs;
 	}
--- a/and-bible/app/src/main/java/net/bible/service/format/osistohtml/taghandler/ReferenceHandler.java
+++ b/and-bible/app/src/main/java/net/bible/service/format/osistohtml/taghandler/ReferenceHandler.java
@@ -135,7 +135,8 @@ public class ReferenceHandler implements OsisTagHandler {
 		        boolean isSimpleContent = content.length()<3 && content.length()>0;
 		        Iterator<VerseRange> it = ref.rangeIterator(RestrictionType.CHAPTER);
 		        
-		        if (isSingleVerse && isSimpleContent) {
+		        // For commentaries and general books, it is necessary to show always reference content. 
+		        if (parameters.isShowReferenceContent() || (isSingleVerse && isSimpleContent)) { // single verse range or single verse
 			        // simple verse no e.g. 1 or 2 preceding the actual verse in TSK
 					result.append("<a href='").append(Constants.BIBLE_PROTOCOL).append(":").append(it.next().getOsisRef()).append("'>");
 					result.append(content);
--- a/and-bible/app/src/main/java/net/bible/service/sword/SwordContentFacade.java
+++ b/and-bible/app/src/main/java/net/bible/service/sword/SwordContentFacade.java
@@ -323,6 +323,11 @@ public class SwordContentFacade {
 			osisToHtmlParameters.setDocumentVersification(((AbstractPassageBook)book).getVersification());
 		}
 		
+		// If Bible, do not show reference content. In commentaries and genbooks they need to be shown
+		if (BookCategory.BIBLE.equals(bookCategory)){
+			osisToHtmlParameters.setShowReferenceContent(false);
+		}
+		
 		if (isAndroid) {
 	    	// HunUj has an error in that refs are not wrapped so automatically add notes around refs
 	    	osisToHtmlParameters.setAutoWrapUnwrappedRefsInNote("HunUj".equals(book.getInitials()));


tuomas2:
Added another small fix to this pull request. Please comment if you see some possible concerns here. I'll push similar fix to jsword to allow spaces in references (if I have understood correctly, OSIS does not prevent spaces in genbook references or book titles)

mjdenham:
Hi Tuomas,

I will need to look at this after the upcoming AB windows functionality
release.  I am a bit stretched at the moment.  It would be good if you also
update the tests for this particular type of reference as the OSIS-html
conversion has reasonable test coverage.

Cheers
Martin

tuomas2:
Okay, no problem.  I'll try to run&fix broken tests (if there are some) at some point. 
ps. You're doing great job with AB windows functionality, I really like it!

mjdenham:
Why should references from GenBooks and Commentaries be treated differently to Bibles?

tuomas2:
Because bible crossreferences are desired to be shown localized form (if I understood correctly from https://groups.google.com/forum/#!msg/and-bible/cSZmdzDU1u0/L9lEaHaCnoQJ) and for bibles this does not really matter, as these link contents are not ever shown in the document (but only if you view crossreferences via menu). In commentaries and genbooks they are desired to be shown as module author has defined (like hyperlinks).  Link content vary a lot, and I would like to see them exactly as in original document. Like I said, I chose this way of implementing by following the example of xulsword, which implements this the same way.

Let me illustrate this with a real example from my case (ESV Study Bible).

Here is screenshot with old version of andbible:
sshot2

And below is the picture of a version that is *fixed* (this PR applied). 

sshot

The document has a lot of bible references and their content formatting varies a lot, depending on the context. Links always point to the first verse (I have made some guessing form the content to get range ends, but it does not work for all cases -- there are *so* many different contents). If those are translated to localized versions (i.e. "Psalm 68:19" instead of "v. 19"), the readibility of the document suffers a lot and also information is lost, if the range end guessing is not correct. 

Hope this explanation helps to understand my real pain with this issue :)

mjdenham:
I am surprised that only the first verse is used for the content and not the entire range.  I was not aware this was happening and I don't think it should be, unless you only have the first verse as the reference.

tuomas2:
Yes, that is the case (I only have the first verse as the reference), like explained above. The reason for my module not having always range end is that my material does not have it available (epubs are practically bunch of HTMLs that have links to first verses -- this target I can translate to osisRef robustly without guessing). 

But don't you agree that hyperlinks should be shown as in original document (here, the same way as in ESV Study Bible paperback - not as localized altered versions)? I mean -- If original says 'vv.17' or whatever suits the context, it should be rendered as 'vv. 17' and nothing else, I think. 

Or if something else is really preferred (which I doubt), it should be optional feature that user can choose what he likes -- on module conf level or in user preferences.

byfaithtogod:
Martin

I agree with tuomas2 - for me problem, when reference has one name, but AB ignores the name of reference...

Emet-like-AndBible:
Martin

I agree with tuomas2 - for me problem, when reference has one name, but AB ignores the name of reference...

mjdenham:
Yes, I can see your point regarding link content and it is particularly problematic in your module because the links are not the same as the link content.  We can give it a try.

Also, we need to consider if allowing spaces will have any consequences.

Does this change in any way depend on you related jsword changes?

Martin

tuomas2:
Yes, consequences need to be surveyed carefully (and BTW, sorry that I did not find time to look at unit tests yet). There might be something that I did not take into account. This change does not in itself depend on jsword changes, but genbook references that contain spaces wont be functional before jsword fix is also applied too (i.e. this should not break anything). 

JSword fix is pretty important I think, it would be great if you would consider including it in And bible test build. On JSword repository it feels somewhat silent these days.

mjdenham:
I defer to DM on issues like that and would rather not risk starting a JSword fork so will probably exclude the space change for now.

mjdenham:
This change does not cope with references that include multiple distinct verses e.g. in Gill Ps 121:1 there are several.

tuomas2:
I think that kind of references are not allowed. So if there are that kind of links in some module, the module should be fixed.

From OSIS manual http://www.crosswire.org/osis/OSIS%202.1.1%20User%20Manual%2006March2006.pdf page 90 chapter 16: 

"A single osisRef cannot identify a discontiguous range of a work. For example, a complex reference such as "John 3:14-16, 18; 4:1-2; 19-20" cannot be encoded as a single reference. It must instead be encoded as several parts, each contiguous".

mjdenham:
But Gill is a very popular module and I am sure there are many others.  It would be interesting to see what Xiphos did.

tuomas2:
So could this module be updated? Where can this Gill module be obtained? I would like to have a look. I haven't seen it on AB lists.

mjdenham:
It is there.  It is hosted in the Xiphos repo.  But I am becoming a little
hesitant regarding this change.  It might be best to concentrate on the zip
import and get that finished first.  I am quite excited about zip import
and can see many uses in hostile countries and environments where a
connection is rarely available.

Regarding scaling your suggestion of maximum 100% sounds a good idea and
hopefully that will be acceptable.

Martin

tuomas2:
May I suggest that we revert only whitespace part for the time being and
return back to it later, if possible. I would appreciate the "inline
reference content" fix in the upcoming release. What do you think?

2015-07-05 0:45 GMT+03:00 Martin <notifications@github.com>:

> It is there. It is hosted in the Xiphos repo. But I am becoming a little
> hesitant regarding this change. It might be best to concentrate on the zip
> import and get that finished first. I am quite excited about zip import
> and can see many uses in hostile countries and environments where a
> connection is rarely available.
>
> Regarding scaling your suggestion of maximum 100% sounds a good idea and
> hopefully that will be acceptable.
>
> Martin
>
> On 4 July 2015 at 17:36, Tuomas Airaksinen <notifications@github.com>
> wrote:
>
> > So could this module be updated? Where can this Gill module be obtained?
> I
> > would like to have a look. I haven't seen it on AB lists.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/mjdenham/and-bible/pull/28#issuecomment-118530748>.
>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mjdenham/and-bible/pull/28#issuecomment-118557038>.
>



-- 
T: Tuomas

tuomas2:
BTW Gill is not OSIS module but ThML. I'm not sure if it is handled with the same code or not, but maybe ThML documents could be allowed to have this kind of formatting.

Emet-like-AndBible:
However there are questions on display of the names of references. still often display full names of biblical books, but not short, as set in the module...

for example, the reference <reference osisRef="Ezek.16.48">Иез. 16:48</reference> when Russian interface displays as "Иезекииль 16:48", and must be short "Иез. 16:48"... or reference <reference osisRef="Ezek.18.30-Ezek.18.31">Ezek 18:30-31</reference> - should appear with the name "Ezek 18:30-31" and appear as "Иезекииль 18:30-31"... id est, I read the English comments, but interface in Russian language, then the names of references in the English comment on Russian, and not in a brief kind... incomprehensible...

tuomas2:
I'd like to get feedback how this should be properly implemented, if this is not acceptable as is. I would like to get this functionality merged at some point.

mjdenham:
We need to be careful when making changes to core code that affects all modules.  Unfortunately this change broke the first module that I tested.  You could attempt to get that module fixed, but also you should conduct tests on a selection of modules to ensure your change does not cause other issues.

Also changes like this require junit tests.

tuomas2:
I separated second commit from this PR to #110. I'll be working on this PR later...

tuomas2:
Because of refactoring I made in #110 I must base my work on that branch. And I need a little different approach anyway. Closing this and opening new PR later. I'll reference this ticket from there.

PR #29, opened 2015-06-25
New feature: Install module from a user-selectable zip file

tuomas2:
This PR adds functionality to select from main menu "Install Documents from a Zip file". This feature is intended to make it much more convenient for users to install propietary modules that cannot be distributed through public repositories.

1. User can give a file through ACTION_GET_CONTENT intent. 
2. Given file is first checked that it contains a configuration file in mods.d and modules directory, and then extracted and registered. 
3. After install is finished, toast message informs the user of the result.


ildar:
Does it install modules into /sdcard/jSword folder? Or the usual
Android/data/... ?

tuomas2:
It installs module in Android/data/...

mjdenham:
Thanks for doing this.  I have just been trying it out and I think some people are going to like it a lot.  I plan to do a test build with it included shortly.

mjdenham:
I have kept this on a separate branch for now but plan to move them over to development later.

mjdenham:
Hi Tuomas, do you think yesterday's changes will solve BV's problem?  Also we should allow reload of modules so a module developer who continuously enhances a module can update easily but there may possibly be jsword issues replacing modules.
Martin

tuomas2:
Sorry, I was not following the mailing list. Now I should get all messages from there. Yes, it seems that BV has faced just the same problem, which I fixed.

mjdenham:
Hi Tuomas,

I notice that you prevent re-installation of modules that already exist.  This might be a nice feature for module developers who wish to try a newer version.  Was there any particular reason you disabled this - like did problems occur when re-installing over an existing module?

Cheers
Martin

tuomas2:
I have uninstalled modules every time before re-installing. The reason that I implemented it this way was that I wanted to make sure that there may not happen unintentional overwriting of files that belong to other modules (if module is bad/broken). Re-installation feature would require some extra checking -- probably best approach would be to first perform uninstall of the module and then install as normally. 

Thanks for pushing this forward ;)

mjdenham:
I can't work out how to change the logic to allow documents to be overwritten.  The flow is a bit different to what I normally use.  If I could work it out then there is still the issue of creating a Book to be uninstalled.
The code to remove an existing module/document/book can be gleaned from DownloadManager:
Books.installed().removeBook(book);

If you have any ideas on the above then you are welcome to try.  If not, the functionality is still very useful even without allowing overwrite.

tuomas2:
Most simple approach would be to inform user of the files that would be overwritten and by his permission proceed (with "if unsure, say No"). What do you think? 

It's okay for me to leave it as is for now also. Those genbook issues that I have reported have bigger priority for me at the moment (I'm might be contributing to them in the (near) future).

mjdenham:
Did you consider using the zip expansion code in JSword - IOUtil.unpackZip()?

tuomas2:
unpackZip uses ZipFile instead of ZipInputStream. ZipFile would have allowed somewhat more convenient handling of zipfile (i.e. for example going through first filenames without compressing whole file, for overwrite checking), but I was not able to get it working with InputStreams (Zip file might not be stored at all in file storage when it comes from GET_ACTION_CONTENT / PICK_FILE intent). Moreover, there are no overwrite checking in unpackZip, so that should have been implemented there.

mjdenham:
It would be good to make Zip file installation accessible as an alternative to Download during initial startup.  This would help in a lot of missionary situations.  I have had this sort of request at various times and again just yesterday.

PR #54, opened 2016-04-20
add saving/restoring history to/from SharedPreferences

cepreu2github:
Now during work every history item saves to SharedPreferences.
At creation of history stack elements reads from SharedPreferences.

### DIFF
--- a/and-bible/app/src/main/java/net/bible/android/view/activity/base/CurrentActivityHolder.java
+++ b/and-bible/app/src/main/java/net/bible/android/view/activity/base/CurrentActivityHolder.java
@@ -40,12 +40,12 @@ public class CurrentActivityHolder {
 	public void iAmNoLongerCurrent(Activity activity) {
 		// if the next activity has not already overwritten my registration 
 		if (currentActivity!=null && currentActivity.equals(activity)) {
-			Log.w(TAG, "Temporarily null current ativity");
-			currentActivity = null;
 			if (appIsInForeground) {
 				appIsInForeground = false;
 				EventBus.getDefault().post(new AppToBackgroundEvent(Position.BACKGROUND));
 			}
+			Log.w(TAG, "Temporarily null current ativity");
+			currentActivity = null;
 		}
 	}
 	
--- a/and-bible/app/src/main/java/net/bible/service/history/HistoryItem.java
+++ b/and-bible/app/src/main/java/net/bible/service/history/HistoryItem.java
@@ -18,4 +18,7 @@ public interface HistoryItem {
 	// do back to the state at this point
 	public abstract void revertTo();
 
+	// if this Item came from SharedPreferences, we shouldn't go to it at Back key pressed
+	public abstract boolean isFromPersistentHistory();
+
 }
--- a/and-bible/app/src/main/java/net/bible/service/history/HistoryItemBase.java
+++ b/and-bible/app/src/main/java/net/bible/service/history/HistoryItemBase.java
@@ -24,4 +24,10 @@ public abstract class HistoryItemBase implements HistoryItem {
 	public Window getScreen() {
 		return window;
 	}
+
+	// only KeyHistoryItem can be serialized/deserialized to SharedPrefs, so return false for all other childs
+	@Override
+	public boolean isFromPersistentHistory() {
+		return false;
+	}
 }
--- a/and-bible/app/src/main/java/net/bible/service/history/HistoryManager.java
+++ b/and-bible/app/src/main/java/net/bible/service/history/HistoryManager.java
@@ -18,9 +18,13 @@ import net.bible.android.view.activity.base.CurrentActivityHolder;
 import net.bible.android.view.activity.page.MainBibleActivity;
 
 import org.crosswire.jsword.book.Book;
+import org.crosswire.jsword.book.Books;
 import org.crosswire.jsword.passage.Key;
+import org.crosswire.jsword.passage.NoSuchKeyException;
 
 import android.app.Activity;
+import android.content.Context;
+import android.content.SharedPreferences;
 import android.util.Log;
 
 /**
@@ -33,6 +37,11 @@ import android.util.Log;
 public class HistoryManager {
 
 	private static int MAX_HISTORY = 80;
+
+	private final static String HISTORY_STR = "history";
+	private final static String KEY_STR = "historyItemKey";
+	private final static String BOOK_STR = "historyItemBook";
+	private final static String TOP_STR = "top";
 	
 	private Map<Window, Stack<HistoryItem>> screenHistoryStackMap = new HashMap<>();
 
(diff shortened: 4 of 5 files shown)


mjdenham:
Thanks for your contribution.  Many people have requested this feature.  I have had a quick look and your approach seems nice and efficient.

We may need to consider multiple windows.  Every window has its own History queue but I think your implementation would replicate the history queue for all windows on restore.

Cheers
Martin

cepreu2github:
I think current behavior is appropriate. I can't even imagine, which rule possibly can regulate how to save or not save history for one or another window. Just save all - it is pretty straightforward and good solution.

cepreu2github:
You can suppose some rule. Maybe I find out how to implement your proposition in code.

cepreu2github:
By the way. This approach is very obvious. Browsers for example just saves history from all tabs and have one global history.

ildar:
Greetings,
having one-window persistent history is the good step forward and may be
already quite useful if it doesn't break current per-window non-persistent
history. Would you consider merging this half-working patch?

mjdenham:
Unfortunately the history restoration would not be predictable and I would get a lot of complaints.

I think that adding windowNumber to the saving might work but am too busy with multi-select changes atm to investigate.

Also, it may be neater to do the save/restore on the AppToBackgroundEvent.

cepreu2github:
Can your provide information, which behavior will be appropriate? In form of some algorithm when and how save history. I can't understand, what the expected behavior, which you consider as good...
Is there any way to find when application goes to background? onPause()/onResume() from Activities it is not appropriate solution because it happens also when user switches from one Activity to another.

mjdenham:
AppToBackgroundEvent is an event triggered by And Bible when the app loses/gains focus.  You will find various examples of its use in the app.

cepreu2github:
Done. Now HistoryManager saves history at AppToBackgroundEvent.
Is there anything else must be done before you can proceed with merging?

ildar:
Сергей, did you consider saving all window's history? Is it complicated?

cepreu2github:
It is already works in such way. You can try it. History are separate for every window, but when you close application and open it again, you will find out that there is one window with one history, but items are gathered together from all previous windows.

ildar:
I didn't look at it,
but it sounds very inconsistent. User expects to find his/her windows in
the way it was before switching from the And Bible. IMHO it is the major
issue. What do you think, Сергей, Martin?

cepreu2github:
As user I prefer application to not restore all windows at startup. I am pretty satisfied with current behavior. What disturbs me is that all my history is losing at every application restart. I prefer to have a way to see, which verses I opened during preaching or counseling without manually adding every verse to bookmarks during session.

ildar:
I see. Am I right that your patch actually does this:
1. Fixes window history to be persistent
2. Vanishes all windows except the 1st
?

cepreu2github:
Nope. Only 1st.
Second point is default behavior without my patch. Application opens one window at startup no matter how many of them was before. So I decided to put history from all previously opened windows to history of this one window. I think is better then loosing history of some windows or all of them (without my patch).

ildar:
Really? I can't reproduce that!
My And Bible always starts with two windows: top and bottom if I left it
that way. All minimized windows are also in place.
What version do you have? Mine is 20160503 (testing)

cepreu2github:
Strange... Tried again with b7e398a2a257a1a6773fc5b33212a8297ce0a942 of master branch. Still same behavior. Only one window at startup.  Tested on emulator (Android 4.4, 480x800 hdpi, x86 ABI).

mjdenham:
And Bible saves and restores the window configuration when moving to and from the background so when And Bible starts it will display the same windows as when it was shut down by Android.

cepreu2github:
Sorry, I wish to make my patch better, but I can't reproduce this behaviour (restoring windows). Therefore I can't understand how to write code to support this feature and how to debug this code later.

mjdenham:
Sorry, but I have never seen it not work.  Have you Ildar?  

What might prevent this working?  You aren't pressing the Back key to reverse out through all the screens are you - some people do that.

ildar:
No, I never seen it misbehaving like that.
Сергей, could you please check the pristine version from github or use
another device or emulator. Also you could run it in Android-x86.

cepreu2github:
[Screencast](http://postimg.org/image/66e6skmaj/) made on android-x86-5.1-rc1.

ildar:
Oh! I got it!
You use non-persistent windows! I don't know the details about the
technical difference between the two though.
To get a persistent windows you go to menu → Windows → New.

mjdenham:
The window opened by pressing on a link is not persisted.

cepreu2github:
Ok, I got it. I disabled persistent history for multi-window with latest commit. But it still works for one-window mode. So there is no confusion for multi-window users because for them all works as before.

cepreu2github:
At least I can't imagine really good way to combine separate (!) temporal (!!) history for one window (!!!) with global (!) persistent (!!) application-wide (!!!) history. Maybe you can suggest some better approach?
I also got idea to get special History dialogue for persistent history separated from usual history. Or in current history mark out items from persistence somehow (colour? icon?) to minimize confusion. But it needs to better discussed before we can implement such feature.

ildar:
Let me express my opinion again.
You can see the example of separate per-window "history" in Firefox for
Android as a per-tab "history". But it actually isn't history, it is the
URL stack with Forward and Back navigation and dropping the tail if a new
URL chosen. But the browser actually also has The History which is just
different.
But I don't feel like a Bible app needs per-window history. Windows usually
are synchronized (except general books) hence just follow the top window's
history, so the state save should be enough for those.
So for me making the topmost window history persistent solves the task if
it doesn't break other things.
This is my pure IMHO.

> --
Ildar Mulyukov,
child of God

email: ildar.mulyukov@gmail.com
GoogleTalk: ildar.mulyukov@gmail.com
blog: http://johan-notes.blogspot.com/

cepreu2github:
You have to decide between yourself, does AndBible need per-window history or not...

cepreu2github:
> I think that adding windowNumber to the saving might work

Because per-window history and persistence is absolutely incompatible things from my point of view. Per-window history is only while window is not closed. What to do with history elements then? Add to history of other window? Sound confusing? Lost them? So why we need persistence at all then if it is norm to us to loose some elements?

> So for me making the topmost window history persistent solves the task

What is the topmost window? How it technically can be determined from other windows?  Because you can close any window. They are same from that perspective.

Therefore only way I see is that I wrote after 46e11ff commit.

tuomas2:
Hi @cepreu2github, could you please resolve merge conflicts by merging current master to your branch and I'll take a look at this again soon and test it in emulator, and give then my thoughts on it.

tuomas2:
Implemented differently in master, so closing.

Issue #30, opened 2015-06-26
Idea: "Open link in current window" from context menu


tuomas2:
Now that there is multiple window support and links from a book/commentary are opened in a special window, in some cases it would be nice to have ability to open link in current window also. 

I think this could be implemented best by using context menu (after long press of a link), from which user could choose function "Open link in current window".

Emet-like-AndBible:
tuomas2

If you make such a possibility, then the logical conclusion is the opposite-if AB preferences globally chosen to open all links in the same window, so that through the context menu, you can open the desired links in a special window.

tuomas2:
This is implemented already.

PR #106, opened 2018-03-08
WIP Make long click reverse option to open in a new window

tuomas2:
Advanced user feature:

When clicking links, it is sometimes conventional to reverse option open link in new temporary window.

By long click it is now possible.

### DIFF
--- a/and-bible/app/src/main/java/net/bible/android/control/link/LinkControl.java
+++ b/and-bible/app/src/main/java/net/bible/android/control/link/LinkControl.java
@@ -55,11 +55,17 @@ public class LinkControl {
 
 	private static final String TAG = "LinkControl";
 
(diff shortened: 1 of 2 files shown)


mjdenham:
Unfortunately, as it is, this will confuse users.

tuomas2:
I kind of agree. This was the simplest thing that came to my mind to implement this feature (as I'm not experienced in android development) and I have found it really useful personally -- I'm using it all the time with commentaries.

I guess some kind of context menu when long-clicking would be better. What do you think, would that be acceptable?

tuomas2:
Related issue ticket #30

tuomas2:
Closing this as #114 is more user friendly approach to this.

PR #107, opened 2018-03-08
Feature / Layout style improvements

tuomas2:
1) for tablets, I like to have some left + right margins.
2) I made a hack which makes it possible to adjust how big gap is between paragraphs. Instead of <br/> we need to write <div class="breakline"> and adjust css accordingly. It is not possible to adjust <br/> by css.

### DIFF
--- a/and-bible/app/src/main/assets/web/style.css
+++ b/and-bible/app/src/main/assets/web/style.css
@@ -29,6 +29,10 @@ span[id="0"].verseNo + h1:after {
 	white-space: pre; 
 }
 
+div.breakline {
+    height: 0.5em;
+}
+
 img.sword {
 	max-width: 100%;
 }
--- a/and-bible/app/src/main/assets/web/style.css
+++ b/and-bible/app/src/main/assets/web/style.css
@@ -4,8 +4,10 @@
 body {
 	line-height: 1.6em; /* repeat this in .verse ; also hard-coded in BibleView*/
 	font-size: 1.1em;
-	margin-left: 2vw; /* vw - Equal to 1% of the width of the initial containing block. */
-	margin-right: 2vw
+	margin: auto;
+	padding-left: 2vw;
+	padding-right: 2vw;
+	max-width: 15cm;
 }
 
 h1, .heading1 {
--- a/and-bible/app/src/main/java/net/bible/service/format/osistohtml/taghandler/DivHandler.java
+++ b/and-bible/app/src/main/java/net/bible/service/format/osistohtml/taghandler/DivHandler.java
@@ -88,7 +88,7 @@ public class DivHandler implements OsisTagHandler {
 	public void end() {
 		DivType type = stack.pop();
 		if (DivType.PARAGRAPH.equals(type) && passageInfo.isAnyTextWritten) {
-			writer.write("<br />");
+			writer.write("<div class='breakline'></div>");
 		} else if (DivType.PREVERSE.equals(type)) {
 			writer.finishInserting();
 		}
--- a/and-bible/app/src/test/java/net/bible/service/format/osistohtml/taghandler/DivHandlerTest.java
+++ b/and-bible/app/src/test/java/net/bible/service/format/osistohtml/taghandler/DivHandlerTest.java
@@ -55,7 +55,7 @@ public class DivHandlerTest {
 		divHandler.start(attrs2);
 		divHandler.end();
 
-		assertThat(htmlTextWriter.getHtml(), equalTo("Some text<br />"));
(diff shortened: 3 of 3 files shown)


tuomas2:
Before:

selection_118

After:

selection_117

tuomas2:
Related: https://andbible.uservoice.com/forums/100177-general/suggestions/1730131-tablet-optimized-interface

mjdenham:
Thanks.

mjdenham:
You have a good point, but I was a bit shocked by the look on my 10" tablet so I added a couple of cm to the maxwidth.  We might have to rethink this for really large devices like televisions and chrome books.

ericjding:
@tuomas2 @mjdenham I don't mind the margins being a little wider on my tablet, but I think this should be a user-configurable setting if possible --when I use my 8" tablet, I don't want to see text displayed as if it's a 6" screen. And if I set the text size to be rather larger, as I prefer to on my tablet, these wide margins take away from, rather than improving, my reading experience.

Having said that, a more serious problem is that this change also appears to make the margins, in my opinion, almost laughably wide in landscape mode on my tablet. Is this intentional?

tuomas2:
Why don't you just flip your tablet to portrait (or use 2 or more windows parallel) and it will be fine? If I use lanscape mode I usually open commentary etc. on another window. My opinion is that wide lines are not very readable.

ericjding:
@tuomas2 Thanks for responding. Of course, you don't personally have to use wide lines if you don't find them readable. That's perfectly fine with me! :-) But others might _prefer_ to use larger screens in landscape mode, so I think this (unexpected?) side effect of your patch is a major and surprising downside for those users. As an aside, I note that I prefer much larger text size on my tablet than you do; perhaps that's why I'm affected by this side effect more than you are...?

To me, it seems that this commit causes new problems, and is thus an undesirable regression in usability, even if the patch makes AndBible work in portrait mode more like you personally prefer it. I'd imagine there will also be other use cases where this side effect would be undesirable too, and just telling people to turn their TVs and ChromeBook screens sideways, for example, isn't going to cut it. :-)

From a usability standpoint, I think a much better choice would be to make side margins user-adjustable, (the way that the Amazon Kindle app does, for example). But even in that case, I think it should be the margins that stay the same between portrait and landscape mode, not the width of the text column!

tuomas2:
I agree that margin size could be user adjustable, as adjustable settings parameter. 
Now it's hard-coded in CSS file.

mjdenham:
I always prefer to avoid ading extra user settings if possible as it makes the app more confusing.  The first step should be to make the margins smaller on very large screens.

ericjding:
Maybe you're just saying the same thing a different way, but shouldn't the first step be to change or remove the hard-coded max-width? That's the reason for the unintended side-effect in landscape mode: the assumption that 15cm is a suitable max-width for all screens and orientations.

Is there a way to modify the CSS rules dynamically, e.g., if it's a tablet, then the max-width is set to 80% or something? I still question the basic assumption that it's a _good idea_ to do this at all by default, and I can't think of any other text-oriented app that I use which does it by default, but if you must do it, then please, at least don't totally cripple landscape display on devices larger than a smartphone and/or which use a large text size for visibility reasons!

ildar:
Greetings,
I see the discussion becoming quite hot ))
I suggest reopen it in forum so other users may participate then come back
here with a decision.
I have some ideas about when margins should be used and when should be
avoided but don't want to spoil this PR

ericjding:
Hmm.... I don't feel hot. :-)  And I don't mean to provoke either @tuomas2 or @mjdenham, I'm just trying to make suggestions for adjusting this UI change so that it suits various configurations better. Surely nobody disagrees that 15cm max-width is a good idea for a widescreen Android TV setup, for example!

In any case, while perusing the source code, I discovered the option to manually override the CSS stylesheet, so I'm using that option for now. Ultimately, I hope that a better way is found to accommodate users who sometimes use And Bible on a larger screen in landscape orientation, but for now, this workaround will have to do.

ildar:
Don't worry, I see this discussion is very appropriate (and polite).
The main criteria for raw width in reading apps is that the whole raw must
be narrow enough for fast reading. But on the other side, A-B is not just a
"reader", sometimes a user wants to review, fast scroll etc. which also an
important use-case. That's why I suggested a discussion with wider audience.
Anyway. It's great you did it with CSS. Can you @ericjding please document
it on the https://github.com/mjdenham/and-bible/wiki ?

ericjding:
@ildar I've added an example to the wiki as you suggested.

You make an interesting case for limiting maximum width, but what is more, what seems narrow (or wide) enough also depends on text size. That is another reason that setting an absolute width in style.css is not an ideal solution, in my opinion.

ildar:
Thank you.

PR #108, opened 2018-03-08
Bugfix / Fix restoring state for Genbooks

tuomas2:
Restoring state of genbooks after app restart is broken. This fixes the problem.


tuomas2:
Please let me know if this requires more explanation etc.

tuomas2:
Example: when osisID of the current document is "Book introductions/Introduction to Exodus/Author", the name is "Author", and it stores the latter into state json.

Then when it tries to restore, it finds multiple instances of "Author" (each book has one) and restores the last matching name, i.e. the one with osisID "Book introductions/Introduction to Philippians/Author and Title". 

This is wrong. Looking up by osisID works always.

PR #109, opened 2018-03-08
Improvement / Show full OSIS ID in genbook content list (instead of name)

tuomas2:
In genbook content list, it is better to show full osis id instead of name, which is only the last part of the osis id. This makes it possible to use content list to get better insight and navigate nested TOC structures.


mjdenham:
Can you give some examples of what is shown and what you think should be shown.

tuomas2:
Before:

selection_114

After:
selection_116

PR #110, opened 2018-03-15
Bugfix / Handle references that contain spaces properly. 

tuomas2:
Spaces are allowed in general book references.This fix was part of https://github.com/mjdenham/and-bible/pull/28. It still had some issues and broken tests. This PR has fixed the issues and I also added test about the broken feature.

Refactored code a little (getReferenceFromContent method)


tuomas2:
This does not break Gill (tested Gill for Ps.121.1 and it displays and cross reference links work as before) and unit tests run correctly. I added a new unit test as well. I think this is ready to be merged, but please review.

mjdenham:
Tuomas, could you summarise for me, the recent changes you have made to the reference handler code as there have been many and I am a little confused regarding this complex area.

One thought that occurred to me regarding the reference code is that it used to display the reference, rather than the contained text, which meant the reference was automatically translated into the user's language.  One book that this was partiularly useful for was TSK as it had very little other text and would consequently act as a list of cross-references for people of any language.

PR #116, opened 2018-03-30
Bugfix / Fix bottom gap in the multiwindow mode

tuomas2:
Since multiwindow support was introduced, bottom gap length has been incorrect if multiple windows are opened. Gap length (getPaddingAtBottom) was calculated from the height of the active window (which can be small and which is set as global screen height) rather than from actual screen height. This fixes it by considering parent view height (view that contains all bible views / windows) height instead of actual bibleview height.

## Before: could not scroll second window down to the bottom 

screenshot_1522397871

## After: scrolling second window is now possible to the bottom

screenshot_1522397931


mjdenham:
I recall thinking about this before, but the correct solution would involve different amounts of padding for each view i.e. more for the larger views and less for the smaller.  Your solution gives all views a lot more padding - always too much.

This seems like a good step forward, but still not perfect.

For now I will just add a review comment regarding null checks.

and-bible/app/src/main/java/net/bible/android/view/activity/page/BibleView.java
```diff
@@ -166,8 +166,9 @@ public boolean onJsAlert(WebView view, String url, String message, JsResult resu
 	@Override
 	protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
 		super.onMeasure(widthMeasureSpec, heightMeasureSpec);
-		// update the height in ScreenSettings
-		ScreenSettings.setContentViewHeightPx(getMeasuredHeight());
+		// update the height in ScreenSettings. Global screen height must be taken from parent view
+		// that contains all windows.
+		ScreenSettings.setContentViewHeightPx(((View) getParent().getParent()).getMeasuredHeight());
```
mjdenham: Do we need null checks here?

tuomas2:
Yes not perfect. Btw one more little fix could be done: I think refreshing is not needed for commentaries and General books at all and causes them to go always to the beginning of the page.

tuomas2:
And it might be good If refreshing would be done after making new bookmarks so that those would be seen also in other windows/languages. I might make another pull request about these at some point

and-bible/app/src/main/java/net/bible/android/view/activity/page/BibleView.java
tuomas2: I agree this is not elegant. About null checks: with the current layout I think they should never be null and that parent of parent element should be correct. What should we do here anyway if the parent happens to be null (i.e. we reach top level element?)? 

If at some point there is a change in layout structure, this might of course break. Thus, I guess some kind of test would indeed be appropriate. But I'm not (yet) familiar how to implement this kind tests for android.

mjdenham: You just compare with null.  Kotlin would be great here.  I can add a couple of null checks.

PR #117, opened 2018-03-30
Bugfix / Fix: special dedicated link window forgets bible version if it is closed

tuomas2:
## Bug description

- Click link in a commentary (gill for example) which opens bible verse in a special window 
- Switch bible version in special window to something else
- Close special window
- Click link in a commentary which opens special window again. 
- Expected: bible verse is opened in a bible version that was last used in special window
- Actual: bible version is opened in a default bible version.

This PR fixes this issue.


and-bible/app/src/main/java/net/bible/android/control/page/window/WindowControl.java
```diff
@@ -116,8 +116,8 @@ public void showLinkUsingDefaultBible(Key key) {
         LinksWindow linksWindow = windowRepository.getDedicatedLinksWindow();
 
         // default either to links window bible or if closed then active window bible 
-		Book defaultBible = linksWindow.getDefaultBible(getActiveWindow());
-        
+		Book defaultBible = linksWindow.getPageManager().getCurrentBible().getCurrentDocument();
```
mjdenham: I think different people would implement this logic in different ways, but that your way is a step forward.

However, one issue is that the first time a link is pressed the first Bible alphabetically will be randomly used, rather than the current Bible (I think).

It might be nice to add extra logic that uses the linksWindow bible if it has been specifically set, but the active window bible if not, but that might get complicated.

Another thing it might be nice to change is to move the logic into getDefaultBible, which might be able to check if a Bible has been selected yet.

I have reservations about my comments because I would not like to add complexity, but what do you think?

tuomas2: I have noticed the same issue and I think the best logic would be as you suggested. If I remember correctly, I was thinking about this when I was implementing this but I did not figure out an easy to way t add this logic because currently there's no way to check if current document is set or not (getCurrentDocument checks if current document is null and if so, gives the first document). Now that I'm thinking about this, perhaps we could add there isCurrentDocumentSet which checks if it is null and then implement logic here using that. I rethink this, maybe tomorrow or next week.

tuomas2: Changed the logic as suggested. But I don't think the logic can be in getDefaultBible as we need to access windowRepository.

PR #118, opened 2018-04-15
Display date bookmark was added in the bookmark list 

jsmitza:
This suggested change shows the date a bookmark was added in the bookmark list. This is one of the changes suggested in [bookmark-improvements](https://github.com/mjdenham/and-bible/pull/111).

Example: See the date on the right below. (Note that the translation is also shown here - this is not added by this pull, but by another pull request)

image

### DIFF
--- a/and-bible/app/src/main/java/net/bible/android/view/util/widget/TwoLine2TitleListItem.java
+++ b/and-bible/app/src/main/java/net/bible/android/view/util/widget/TwoLine2TitleListItem.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright (C) 2006 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


tuomas2:
Please change destination branch to development, here and in all your pull requests.

and-bible/app/src/main/java/net/bible/android/view/util/widget/TwoLine2TitleListItem.java
```diff
@@ -0,0 +1,97 @@
+/*
```
tuomas2: To me it looks like this class could be done more simply by inheriting from TwoLineListItem.  Or what do you think?

mjdenham:
Oops, I noticed Tuomas comment about the target branch just after I pressed merge.  We normally merge new changes to development.

Had to revert, then couldn't cherry-pick, However, I have manually copied the files over.

PR #119, opened 2018-04-15
Allows refreshing of an existing bookmark's date to the current time when reselected.

jsmitza:
We do this by showing the Add button also on selected verse even if it was bookmarked already. Adding it again will update the bookmark's date to the current time.

Example below. (The star indicates existing bookmark on this verse; the blue highlight indicates the verse was selected again - and the Add button is still shown at the top).

image


mjdenham:
i changed this around a bit after merging as I realised that labels would become disconnected after refresh.

PR #120, opened 2018-04-15
Allow quick selection of many bookmarks to label or delete.

jsmitza:
A long click on a bookmark in the bookmark list will select all bookmarks between the current and previous (long click selected) bookmark.

Clarification (by @timbze 2019-10-09)
And Bible can select multiple bookmarks: long press 1 bookmark and then single tap each other bookmark you also want to select.
This PR is that you can long press one and instead of tapping each one you want to select just long press, or activate with button to single press another one further down the list to select all bookmarks in between as well.


mjdenham:
This ui interaction does not seem obvious to the user and so I am afraid this will not be used as it is.

jsmitza:
I can understand the need for a better UI to do this.   There is a need for something like this though. If you have 2-3 years of bookmarks (like me) and you want to delete some of the 600+ entries,  you have to individually select each bookmark - which is not really feasible. (My thought was that even if not documented this feature at least gives a way to do that :-))

tuomas2:
Hi @jsmitza could you please resolve merge conflicts by merging current master to your branch and I'll take a look at this again soon? I'd like to test this in emulator and give my thoughts.

jsmitza:
Ok merge is done from master to this branch. Actually it is only one file (bookmarks) that affects the change suggested by this branch.  Runs successfully on my box. 

Usage again: Normally a long press on bookmarks will enter selection mode and you can toggle selection by clicking on any bookmark. The selected bookmarks can then be deleted or labelled. NEW: this change support long-press in selection mode as well - it will auto-select ALL bookmarks between the last two long-presses. So it is easy to select many entries to label/delete.

tuomas2:
My suggestion is that you would implement "Select range" button (with a nice icon instead of red marking in the picture below) when selection mode is enabled, and if this "select range" mode is enabled, then you could select ranges by normal click (rather than long click).

Then this feature would be obvious to users. 

What do you think?

screenshot_1544863222

jsmitza:
That sounds good to me and will probably be more user friendly. Actually almost anything that just adds the functionality will be good to me :-) If you want me to do the change it will have to wait a week or three for my holiday.

tuomas2:
Take your time :)

jsmitza:
Hi, same apply here as with the other pull request: I only want to go through that effort if you do plan to incorporate this feature. If not it is not worth it. So bottom line: Do you intend to use this if I change it (in 2 weeks time)?

tuomas2:
> Hi, same apply here as with the other pull request: I only want to go through that effort if you do plan to incorporate this feature. If not it is not worth it. So bottom line: Do you intend to use this if I change it (in 2 weeks time)?

Yes this can be merged if you follow my suggestions.

jsmitza:
Actually there still might be confusion with this suggestion:
1) If you click on the "strawberry" icon you get into the range-selection mode. How do you get out of it?
2) To get into single-selection mode from the normal bookmark list, you have to long-press an item (i.e. there is no icon at the top to select this mode - which users may look for).

So I am still not sure what would be ok. 

Another alternative is - if you keep my original suggestion but add a help message (as a flash alert, or staying open?) which says something like - "Click for single select, long-click for range select".

So I have not made any changes (yet), except for merging to the latest master.

ildar:
Hi! Sorry I'm late to this discussion.
I think that the "strawberry" way is questionable. The button attracts
user's attention. But then it should be easy to use and understandable. UX
is important.
The way of "the hidden feature" (long tap then again long tap): I like it
better: if someone needs it and asks, you can give one sentence explanation
and viola...

jsmitza:
If it is a toggle-button, then the "range-selection-off" mode is actually indicating single-select mode. So you could have this icon from the start on the normal booklist. Then the long press is not really necessary.

The above would be giving all the options via icons; idar's comments (and my original) would be the minimal (hidden) approach.

jsmitza:
Regarding icons (both for this request and for the show-books request): I am not the one to design or find the appropriate icon that is free to use. Somebody will need to help please.

ildar:
On quitting multi select: Google's pattern is: when an item selected then
toolbar is changed in color and elements. The toolbar has the arrow to the
right in top left corner that gets user it of the Select mode. E.g see
Google Inbox

PR #121, opened 2018-04-15
Improve current book/chapter highlight in selection grid

jsmitza:
Improve the highlight of the current book or chapter in the book and chapter selection grid by underlining the name/chapter. (Why: It is difficult to see the current item otherwise, and it helps to quickly locate the current book on the grid. If you re-select it you can change the chapter.)

Examples below:
image

and

image


ildar:
Did you play with changing of background?

jsmitza:
Yes (that would be preferred!), but I could not figure out how to change the button background within the time I had. Wish somebody could do that.

PR #112, opened 2018-03-28
Feature/Add new split mode preference (vertical, horizontal, automatic) to preferences screen

tuomas2:
Currently split mode is determined by if the mobile device is used in portrait or landscape mode.
Sometimes it is more desirable to manually select desired split mode to be vertical even in portrait mode or horizontal in landscape mode. This pull request implements the new preference which allows user to chooce desired behavior. Default is 'automatic', which corresponds to current behavior.

## Demo screenshots. Split mode set to 'vertical':
screenshot_1522235780

## New split mode preference:
screenshot_1522235794

## Split mode preference selection:
screenshot_1522235800

PR #127, opened 2018-04-29
Switch off current window syncing when opening "Open link here" from context menu

tuomas2:
Per user testing / request in #112 .

### DIFF
--- a/and-bible/app/src/main/java/net/bible/android/control/link/LinkControl.java
+++ b/and-bible/app/src/main/java/net/bible/android/control/link/LinkControl.java
@@ -297,6 +297,13 @@ public class LinkControl {
 
 			windowControl.addNewWindow(document, key);
 		}
+		else if(windowMode.equals(WINDOW_MODE_THIS)) {
+			if (document==null) {
+				document = defaultDocument;
+			}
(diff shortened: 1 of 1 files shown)


and-bible/app/src/main/java/net/bible/android/control/link/LinkControl.java
```diff
@@ -297,6 +297,13 @@ else if(windowMode.equals(WINDOW_MODE_NEW)) {
 
 			windowControl.addNewWindow(document, key);
 		}
+		else if(windowMode.equals(WINDOW_MODE_THIS)) {
+			if (document==null) {
+				document = defaultDocument;
+			}
+			windowControl.getActiveWindow().setSynchronised(false);
```
mjdenham: I realise this was suggested but switching off sync is potentially confusing because it is not obvious to the user why this has happened.  I think I preferred the previous issue of other windows being sync'd - they can always press Back.

tuomas2: I agree that this is disputable. But in a way I found it something that might be often desired in typical use case.

tuomas2:
I'll close this for now as we aren't going to merge this

PR #130, opened 2018-04-29
Bugfix/app restarts after assigning bookmark label

tuomas2:
Bug description:
- go to settings and change application language to "English" (or anything else but not "Default").
- add bookmark
- click "Assign labels" button that appears in the bottom (before it disappears)
- click OK
-> Expected: return to main screen without restart
-> Actual: App restarts unnecessarily.

This PR fixes this issue. This does not seem regression but older issue.
When debugging this I found also 2 lines of code which did not make any sense (commit 
c92e18b), but did not cause any issue either. I cleaned it up a little and checked that it did not break.


tuomas2:
Issue fixed better in #131.

PR #131, opened 2018-05-01
Fix window synchronization related problems

tuomas2:
There was regression (from 9b40cf362a840193cd302365505f2bad3507ccae) that caused issue fixed by #122 reappear. This fixes that again and tries to make window synchronization code more clear.

- Rename BookmarkChangedEvent to SynchronizeWindowsEvent which can be triggered when window sync is needed.

- Create new request code RESTART_REQUIRED to IntentHelper for settings window changes.

### DIFF
--- a/and-bible/app/src/main/java/net/bible/android/control/page/window/WindowControl.java
+++ b/and-bible/app/src/main/java/net/bible/android/control/page/window/WindowControl.java
@@ -8,7 +8,7 @@ import net.bible.android.BibleApplication;
 import net.bible.android.activity.R;
 import net.bible.android.control.ApplicationScope;
 import net.bible.android.control.event.EventManager;
-import net.bible.android.control.event.passage.BookmarkChangedEvent;
+import net.bible.android.control.event.passage.SynchronizeWindowsEvent;
 import net.bible.android.control.event.passage.CurrentVerseChangedEvent;
 import net.bible.android.control.event.window.NumberOfWindowsChangedEvent;
 import net.bible.android.control.event.window.WindowSizeChangedEvent;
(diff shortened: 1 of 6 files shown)


and-bible/app/src/main/java/net/bible/android/view/activity/page/MenuCommandHandler.java
```diff
@@ -89,7 +89,7 @@ public boolean handleMenuRequest(MenuItem menuItem) {
 		        	handlerIntent = new Intent(callingActivity, SettingsActivity.class);
 		        	// force the bible view to be refreshed after returning from settings screen because notes, verses, etc. may be switched on or off
 		        	mPrevLocalePref = CommonUtils.getLocalePref();
-		        	requestCode = IntentHelper.REFRESH_DISPLAY_ON_FINISH;
+		        	requestCode = IntentHelper.RESTART_REQUIRED;
```
mjdenham: I don't think this is correct now.  REFRESH_DISPLAY_ON_FINISH is used when settings may have changed.  If the locale setting is changed then a restart is also required.  I suspect this will not cause a refresh for non locale changes e.g. change theme or show footnotes.  At bestthe name is now wrong.

and-bible/app/src/main/java/net/bible/android/control/page/window/WindowControl.java
```diff
@@ -300,7 +300,7 @@ public void onEvent(CurrentVerseChangedEvent event) {
 		windowSync.synchronizeScreens();
 	}
 
-	public void onEvent(BookmarkChangedEvent event) {
+	public void onEvent(SynchronizeWindowsEvent event) {
```
mjdenham: I am confused BookmarksChanged should cause a screen refresh not a windows sync.

tuomas2: Perhaps better name would be something like "ScreenRefreshEvent"?

PR #145, opened 2018-09-03
Allow dictionary key matches to be more flexible.

huperniketes:
Allow user to enter search terms that don't include leading characters (eg, without "00" in Strong's).


mjdenham:
I believe that matching anywhere might be good for some dictionaries but not others.  Which documents have you tried this out on?

huperniketes:
I made the change so I could enter numbers at a slightly faster clip in the Strong's dictionaries. Which directories do you think might be problematic? I'll test them and provide a workaround should they fail to work similarly to the Strong's.

tuomas2:
I haven't been using this dictionary lookup feature personally but this change is making sense to me. I was just testing now with "American tract society bible" as well as Strong's, and in both I think this would make sense. We could try out in next beta and revert if there are negative implications.

PR #150, opened 2018-11-16
Change tabs to whitespaces so that codebase is consistent

tuomas2:


tuomas2:
@mjdenham Mixed tabs + whitespace indentations causes often problems when performing code reviews etc. So I would like to get codebase consistent. Can you think of any reason why I should not do this? As we don't have any large pending pull requests this should not cause big problems to anyone, I guess.

tuomas2:
No need to go diff through, this was done automatically like this:
``` 
find . -name \*.java -type f -exec sed -i.orig 's/\t/    /g' {} +
find . -name \*.kt -type f -exec sed -i.orig 's/\t/    /g' {} +
find . -name \*.xml -type f -exec sed -i.orig 's/\t/    /g' {} +
```

PR #151, opened 2018-11-16
Change whitespace indentations to tabs

tuomas2:
Alternative solution to #150


tuomas2:
I'll do this later when there's no PR's

PR #259, opened 2019-01-29
set View.SYSTEM_UI_FLAG_LAYOUT_STABLE flag when enabling fullscreen mode

ericjding:
In hideSystemUI/showSystemUi, systemUiVisibility was previously set without View.SYSTEM_UI_FLAG_LAYOUT_STABLE (it was commented out, but I'm not sure why). The result is that when AndBible goes into fullscreen mode, the navigation bar background disappears immediately, and the remaining icons then slide downward, which is visually distracting and unattractive.

Setting the View.SYSTEM_UI_FLAG_LAYOUT_STABLE flag results in Android making the navigation bar background and icons slide downward and fade out together instead. I also did a little code refactoring to make the code a little easier to read and understand (in my opinion).


tuomas2:
Seems to work. Good work!

Issue #265, opened 2019-01-31
Reading Plan; allow user to tap checkmark to complete chapter


timbze:
In the Reading Plan page, you can click the Passage button then press back, and the check turns green. I am reading the Bible in my paperback Bible, but using the app to keep track of my plan, so it would be easier for me to just press the check and have it completed.
**I already wrote the code** for myself, very simple fix, but I don't know if this is something that you want in the app as a feature. I will add it if you would like.

PR #266, opened 2019-02-01
Issue #265 Allow user to tap checkmark to complete chapter in reading plan

timbze:
Here's the code to allow the user of Bible reading plan to tap the checkmark in order to mark the chapter as completed.
My .gitignore file is included, I had to make these changes so that those files did not show up in my Windows computer, but don't know if these changes should be public .gitignore file.


app/src/main/java/net/bible/android/view/activity/readingplan/DailyReading.java
```diff
@@ -126,9 +126,19 @@ public void onCreate(Bundle savedInstanceState) {
...
+				mImageTickList.add(mImageTick);
+				// Allow check box to be clicked to mark off the day
+				mImageTick.setOnClickListener(new OnClickListener() {
+					@Override
+					public void onClick(View view) {
+						ReadingStatus status = readingPlanControl.getReadingStatus(mDay);
+						status.setRead(readingNo);
+						updateTicksAndDone();
+					}
+				});
```
tuomas2: Hmm.. This pull request reveals that there's still some mixed tabs/spaces in the code. I need to check this. Problem is not in your code but old codebase.

timbze: I think I saw a note in Git somewhere about changing LF line endings to CRLF. I don't find that note now. I don't know if it would have anything to do with that.

timbze: I see now what you mean. There are a mixture of spaces and tabs in this file. It is very easy to fix in Android studio. I can resubmit a pull request with the fix if you want. Are you using spaces?

tuomas2: I think it's better that you do fix that yourself in this pull request as your pull request is OK (i.e. tabs are used consistently).

I'll go through all codebase over again with some automated method as a separate task. (I did it already once, but for some reason I have missed some...)

timbze: Sounds good. For future do I need to have my editor defaulting to tabs or spaces? Or does it matter?

tuomas2: I'm using .editorconfig, and Android Studio obeys it, so you should be set up OK.

timbze:
Mistakenly added more commits then I had wanted. Closing pull request. Will open properly.

PR #267, opened 2019-02-01
Bible reading plan updates

timbze:
Added a new Bible Reading plan. (Here's the info on the plan. https://www.biblestudytools.com/bible-reading-plan/prof-horners-reading-system.html)

Also modified the reading plan screen, added Next Day button. And added ScrollView so that the plan can have 10 chapters and the list can be scrolled.
Prior to this update, the Done button would go to next day in reading plan if you are not caught up with the reading, and if you were up to date, then exit out of the screen. Now the Done button exits back to Bible view, and the Next Day button always takes you to next day.
I may still add something in the title so it's easier to see if you are up to date or how far behind, but my opinion is that this is better like this than it was. I welcome any other suggestions.


timbze:
Closing this request to submit another one properly.

PR #276, opened 2019-02-04
Reading plan Prof. Horner 10-chapter

timbze:
This reading plan will be useless without the fix in PR #275.
Added 10-chapter-a-day reading plan. More info on the plan here: https://www.biblestudytools.com/bible-reading-plan/prof-horners-reading-system.html


timbze:
Another unexpected commit came along with this. Closing to open correctly.

PR #277, opened 2019-02-04
Reading plan Prof. Horner 10-chapter

timbze:
This reading plan will be useless without the fix in PR #275.
Added 10-chapter-a-day reading plan. More info on the plan here: https://www.biblestudytools.com/bible-reading-plan/prof-horners-reading-system.html


tuomas2:
Do we have rights to distribute this? Can you check?

timbze:
I don't find a license, I think it is free to distribute, but I sent an email to the person who set it up. Waiting for a response.

tuomas2:
We have permission to distribute this (given by email from Prof. Horner himself), so merging..

Issue #270, opened 2019-02-03
Android back button behavior when main menu is active.... 


bvahedy:
When the main menu is displayed, hitting the Android back button should close the menu instead it reverts back to the item previous in the history.
Note that this works correctly with the Bible menu window on the top right of the menu bar but not the menu option on the top left.

timbze:
@bvahedy @tuomas2 

There may be a different underlying issue, but I tested that it can easily be fixed by adding override fun onBackPressed in MainBibleActivity.kt

tuomas2:
You are right. Feel free to make a fix if you want @timbze :)
Fix will be released in 3.1

timbze:
Fixed in PR #279

bvahedy:
The back button correctly closes the menu now, but we need to disable long press on the back button when menu is in display. So, I'll reopen this issue as it causes the history to show but selecting any item in history doesn't do anything.
Note: that the right menu on the menu bar behaves correctly.

tuomas2:
Good catch. I didn't even know about this long-press-back button feature :)

timbze:
Should be a simple fix I think. I can do it tomorrow if it's not done yet by then :)

tuomas2:
I'll leave this to you then 😊

bvahedy:
Yes, I believe Tim is the one who implemented it. It's a nice feature.

timbze:
Maybe Tim implemented it but not this Tim. :)
Submitting a fix now

bvahedy:
When I was growing up and things went South, we always blamed Tim 😁
It's a great mod, just needs a bit of tweaking. Thanks for enhancing it
further.

tuomas2:
Fix will be in 299.

Original feature was most likely by Martin (before project lead role was transferred to me).

bvahedy:
Verified fix in build 315.

PR #279, opened 2019-02-04
Fix bug #270 (back button on DrawerLayout)

timbze:
Fix bug #270 that back button does not close DrawerLayout (main menu on left)


tuomas2:
Good work, thanks!

PR #288, opened 2019-02-08
Checkmark toggle fix Reading plan for #265

timbze:
The code to allow user to toggle checkbox in Reading Plan.

### DIFF
--- a/app/src/main/java/net/bible/android/control/readingplan/ReadingStatus.kt
+++ b/app/src/main/java/net/bible/android/control/readingplan/ReadingStatus.kt
@@ -55,6 +55,11 @@ open class ReadingStatus(
         saveStatus()
     }
 
+    open fun setUnread(readingNo: Int) {
+        status.set(readingNo, false)
+        saveStatus()
+    }
+
     open fun isRead(readingNo: Int): Boolean {
         return status.get(readingNo)
     }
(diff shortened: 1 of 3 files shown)


app/src/main/java/net/bible/android/control/readingplan/ReadingStatus.kt
```diff
@@ -55,6 +55,11 @@ open class ReadingStatus(
         saveStatus()
     }
 
+    open fun setUnread(readingNo: Int) {
```
tuomas2: Why open? Open is needed only if you inherit this class and want to override function.

timbze: I wasn't sure why but didn't dig in to see. Now I see why. The idea seems to have been to use HistoricReadingStatus to inherit ReadingStatus, but as readonly, therefore the override. I modified HistoricReadingStatus now to also add the setUnread as override for now. Would it be an idea to accept for now in this state of the program till I get to my changes? There will be many major implementation changes with what is in my plans. The HistoricReadingStatus will probably get dropped.

tuomas2: Ok!

PR #315, opened 2019-02-16
Long back press when menu drawer is open #270

### LINKED ISSUE #270: Android back button behavior when main menu is active....

timbze:
Closes #270 
Long back press shouldn't bring up history screen when menu drawer is open

### DIFF
--- a/app/src/main/java/net/bible/android/view/activity/page/MainBibleActivity.kt
+++ b/app/src/main/java/net/bible/android/view/activity/page/MainBibleActivity.kt
@@ -256,6 +256,14 @@ class MainBibleActivity : CustomTitlebarActivityBase(), VerseActionModeMediator.
         }
     }
 
+    override fun onKeyLongPress(keyCode: Int, event: KeyEvent): Boolean {
+        if (drawerLayout.isDrawerVisible(GravityCompat.START) &&
+            keyCode == KeyEvent.KEYCODE_BACK)
+            return true
+
+        return super.onKeyLongPress(keyCode, event)
+    }
+
     private fun setupToolbarButtons() {
         homeButton.setOnClickListener {
             if(drawerLayout.isDrawerVisible(GravityCompat.START)) {


app/src/main/java/net/bible/android/view/activity/page/MainBibleActivity.kt
```diff
@@ -256,6 +256,14 @@ class MainBibleActivity : CustomTitlebarActivityBase(), VerseActionModeMediator.
         }
     }
 
+    override fun onKeyLongPress(keyCode: Int, event: KeyEvent): Boolean {
+        if (drawerLayout.isDrawerVisible(GravityCompat.START) &&
+            keyCode == KeyEvent.KEYCODE_BACK)
+            return true
```
tuomas2: I think you should return false here.

timbze: Could you explain why?
I thought the idea is that I have handled it and don't want anyone else to handle it, I return true. Maybe I don't know something or am not thinking about something. 
Here's android's docs that I thought is my idea, but again, maybe I'm not seeing something.
https://developer.android.com/reference/android/view/KeyEvent.Callback.html#onKeyLongPress(int,%20android.view.KeyEvent)
Returns boolean | If you handled the event, return true. If you want to allow the event to be handled by the next receiver, return false.

tuomas2: But there's no handling for it here so by this logic it should return false. 
On the other hand, if some action is taken, then it should return true.

But this is a bit unclear case indeed. If this is to say "do not ever do anything here, prevent any possible default actions by Android system etc." then true indeed is right.

tuomas2: Here we essentially want to prevent what is happening in super call.
BTW, I think the code from ActivityBase from lines 152-158 should be moved here. 
Could you check / do that too?

timbze: > But this is a bit unclear case indeed. If this is to say "do not ever do anything here, prevent any possible default actions by Android system etc / other (child?) components etc." then true indeed is right.
That was my reasoning. 
In ActivityBase onKeyLongPress it seems to be return true in these cases.
I guess you just say the word, should we return true or false?

> BTW, I think the code from ActivityBase from lines 152-158 should be moved here.
Could you check / do that too if you agree with me?
Yes I think you're right.

PR #329, opened 2019-02-26
Update play store entries

tuomas2:
These need to be manually updated to play store.


tuomas2:
Ukrainian updated. (not others yet)

Issue #148, opened 2018-11-16
Back up (backup) bookmarks and notes to/from Google Drive, manual via intents


tuomas2:
In addition, it might be possible to implement simple synchronization mechanism with google drive too.

tuomas2:
Manual backuping is implemented in build 306. @bvahedy could you test this?

bvahedy:
This send to be working fine in 307 beta.

tuomas2:
Implemented in 308

PR #348, opened 2019-04-18
Implement backup / restore via intent

tuomas2:
Implement #148 (manual only)

### DIFF
--- a/app/src/main/java/net/bible/android/control/backup/BackupControl.kt
+++ b/app/src/main/java/net/bible/android/control/backup/BackupControl.kt
@@ -18,6 +18,7 @@
 
 package net.bible.android.control.backup
 
+import android.app.Activity
 import android.os.Environment
 import android.util.Log
 import android.view.Menu
@@ -32,9 +33,17 @@ import net.bible.android.view.activity.base.Dialogs
 import net.bible.service.common.FileManager
 import net.bible.service.db.CommonDatabaseHelper
 
+import android.content.Intent
+import android.net.Uri
+import androidx.core.content.FileProvider
+import kotlinx.serialization.internal.readExactNBytes
+import net.bible.android.activity.BuildConfig
 import java.io.File
-
+import java.io.FileOutputStream
+import java.io.InputStream
+import java.io.OutputStreamWriter
 import javax.inject.Inject
+
 
 /**
  * Support backup and restore of the And bible database which contains bookmarks and notes.
@@ -42,7 +51,7 @@ import javax.inject.Inject
  * @author Martin Denham [mjdenham at gmail dot com]
  */
 @ApplicationScope
-class BackupControl {
+class BackupControl @Inject constructor() {
 
     /** return true if a backup has been done and the file is on the sd card.
      */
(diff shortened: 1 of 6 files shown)


app/src/main/java/net/bible/android/control/backup/BackupControl.kt
```diff
@@ -67,6 +76,55 @@ class BackupControl {
         }
     }
 
+    /** backup database to custom target (email, drive etc.)
+     */
+    fun backupDatabaseViaIntent(callingActivity: Activity) {
+        val fileName = CommonDatabaseHelper.DATABASE_NAME
+        internalDbBackupDir.mkdirs()
+        FileManager.copyFile(CommonDatabaseHelper.DATABASE_NAME, internalDbDir, internalDbBackupDir)
```
timbze: In Android 9, Sqlite has changed to now use a .db-shm and .db-wal file on top of the .db file. More info here: https://stackoverflow.com/a/54056779/877453 -- As this answer says, a PRAGMA checkpoint must first be run so that the DB file will have all the changes. I found this out while trying to look into the db file while working on the reading plan and not finding the changes that I knew should be there.

That means, also on a restore, the existing db-shm and db-wal files would need to be removed, probably.

I haven't tested this with backing up, but I'm fairly certain this will have to be addressed for Android Pie.

app/src/main/res/values/strings.xml
```diff
@@ -419,4 +419,9 @@ TODO: error and download_document_confirm_prefix need to be parameterized rather
     <!-- Notifications -->
     <string name="notification_channel_tts_status">Text to speech</string>
     <string name="notification_channel_progress_status">Progress status notification for document downloads etc.</string>
+    <string name="backup_email_subject">And Bible database backup</string>
+    <string name="backup_email_message">Your And Bible database containing bookmarks and notes is attached</string>
+    <string name="backup_storage">Backup to storage</string>
+    <string name="restore_storage">Restore from storage</string>
+    <string name="restore_unsuccessfull">Restoring database was unsuccessfull. Perhaps you gave invalid database file.</string>
```
timbze: Spelling suggestion: unsuccessful
`<string name="restore_unsuccessful">Restoring database was unsuccessful. Perhaps you gave invalid database file.</string>`

app/src/main/java/net/bible/android/control/backup/BackupControl.kt
```diff
@@ -94,6 +152,7 @@ class BackupControl {
 
         // this is now unused because And Bible databases are held on the SD card to facilitate easier backup by file copy
```
timbze: Out of date comment?

app/src/main/java/net/bible/android/control/backup/BackupControl.kt
```diff
@@ -67,6 +76,55 @@ class BackupControl {
...
+        val fileName = CommonDatabaseHelper.DATABASE_NAME
+        internalDbBackupDir.mkdirs()
+        val f = File(internalDbBackupDir, fileName)
+        var ok = false
+        val header = inputStream.readExactNBytes(16)
+        if(String(header) == "SQLite format 3\u0000") {
+            val out = FileOutputStream(f)
+            out.write(header)
+            out.write(inputStream.readBytes())
+            out.close()
```
timbze: I guess writing the file via input stream is always safe? Somehow I feel uneasy about that, but I guess it should not be an issue.

app/src/main/res/values/strings.xml
```diff
@@ -378,7 +378,7 @@ TODO: error and download_document_confirm_prefix need to be parameterized rather
     <string name="backup_success">My Notes and Bookmarks saved in %s on the SD card</string>    
     <string name="restore">Restore</string>
```
timbze: Maybe menu item should be "Restore...", and likewise "Backup...", to indicate that another screen will come up?

Also, though not directly related to this PR, I think the user should be asked to confirm that he/she wants to restore and replace his current database (which will include notes, bookmarks, and reading plans)

app/src/main/java/net/bible/android/control/backup/BackupControl.kt
tuomas2: Good point. Need to add this to old-style backupDatabase too.

app/src/main/res/values/strings.xml
tuomas2: Fixed

app/src/main/java/net/bible/android/control/backup/BackupControl.kt
tuomas2: Fixed

app/src/main/java/net/bible/android/control/backup/BackupControl.kt
tuomas2: Why not? How would you do this?
I'm reading from inputstream and writing into FileOutputStream.

app/src/main/res/values/strings.xml
tuomas2: Fixed both

app/src/main/java/net/bible/android/control/backup/BackupControl.kt
timbze: I was thinking of handling the Intent data as a file instead of input stream, but these are just my "feelings" per se. Don't have much experience on Android with either one of them.

tuomas2:
I think this is the only way you can do it. File content can come from Any
Source not Just from File System.  like Google Drive for example

app/src/main/java/net/bible/android/control/backup/BackupControl.kt
tuomas2: I think this is the only way you can do it. File content can come from any
source, not just from file system. Like Google Drive for example.

timbze: Ok

tuomas2: So this way you need to implement reading plan importing in you PR :)

timbze: Already started working on it :)
I only use the inputStream anyway so haven't tried with file object yet.

PR #350, opened 2019-04-18
Fade in/out window buttons instead of slide-in/out

### LINKED ISSUE #349: Change window button animations to fade-in/out
This might work better than current slide-in/out. Also, after fading out, we could perhaps leave window button visible but a lot more transparent.

tuomas2:
Closes #349

### DIFF
--- a/app/src/main/java/net/bible/android/view/activity/page/screen/DocumentWebViewBuilder.kt
+++ b/app/src/main/java/net/bible/android/view/activity/page/screen/DocumentWebViewBuilder.kt
@@ -327,7 +327,7 @@ class DocumentWebViewBuilder @Inject constructor(
         toggleWindowButtonVisibility(true)
         timerTask?.cancel()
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/net/bible/android/view/activity/page/screen/DocumentWebViewBuilder.kt
```diff
@@ -327,7 +327,7 @@ class DocumentWebViewBuilder @Inject constructor(
         toggleWindowButtonVisibility(true)
         timerTask?.cancel()
 
-        if(mainBibleActivity.fullScreen) {
+        if(true || mainBibleActivity.fullScreen) {
```
timbze: Something temporary? true will always pass as true :)

tuomas2: Temporary indeed :-) Good catch, forgot this that way. I'll remove if block.

PR #355, opened 2019-04-25
Feature/#245 tabs

### LINKED ISSUE #245: Tabs for multiple complete bible screen setups
Implement chrome-like tabs, that can contain whole bible screen (including all their splitted windows) setups.

tuomas2:
Closes #245


app/src/main/java/net/bible/android/view/activity/page/MainBibleActivity.kt
```diff
@@ -529,6 +529,8 @@ class MainBibleActivity : CustomTitlebarActivityBase(), VerseActionModeMediator.
 
     private fun chooseTab() {
         val tabs = tabStrings
+        if(tabs.size < 2) return
```
timbze: I guess your haveTabs property would also do the same thing. Doesn't matter, this works too.

PR #362, opened 2019-05-12
Improvements to maximize mode

### LINKED ISSUE #353: Improve maximize mode behavior
In maximized mode:
- show all other windows as "minimized" windows. 
- Clicking "minimized" windows, should switch that particular window to maximized mode

That would allow easy flipping between windows if you want to work in maximized mode.

### LINKED ISSUE #364: App crash after download documents screen (Alpha)
App crashes after returning to Bible screen from download documents. Happens only after 305 (bug was introduced here 24ed4c5496d1e0a285a2016b4262c90640eb0a1f)

From main Bible screen, open download documents
Press "up" arrow on top left to go back.
App crashes

timbze:
Closes #353 
Closes #364
In maximized mode:
- show all other windows as "minimized" windows.
- Clicking "minimized" windows, switches that particular window to maximized mode
- To get out of maximized mode, click the 3-line button on top right and uncheck "Maximize" box.

Allows easy flipping between windows if you want to work in maximized mode.

### DIFF
--- a/app/src/main/java/net/bible/android/control/page/window/WindowControl.kt
+++ b/app/src/main/java/net/bible/android/control/page/window/WindowControl.kt
@@ -288,17 +288,23 @@ open class WindowControl @Inject constructor(
     }
 
     fun restoreWindow(window: Window) {
+        if (window == activeWindow) return
+
         var switchingMaximised = false
         windowRepository.maximisedScreens.forEach {
             switchingMaximised = true
             it.windowLayout.state = WindowState.MINIMISED
         }
-        
-        window.windowLayout.state = if (switchingMaximised) WindowState.MAXIMISED
-        else WindowState.SPLIT
+
+        window.isMaximised = switchingMaximised
 
         // causes BibleViews to be created and laid out
         eventManager.post(NumberOfWindowsChangedEvent(windowChapterVerseMap))
+
+        if (switchingMaximised) {
+            activeWindow = window
+            if (!activeWindow.initialized) PassageChangeMediator.getInstance().forcePageUpdate()
+        }
 
         windowSync.setResynchRequired(true)
         windowSync.synchronizeScreens()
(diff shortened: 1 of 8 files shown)


timbze:
@tuomas2 Can you check why 1 test is failing? It seems to be textProgressionFinPR in SpeakTests.kt and I can't figure out how any of the changes in this PR affects that test. It passes fine in my Android Studio.

tuomas2:
Tested this in emulator.

Issues: 
- When in maximize mode, un-maximize icon should be shown instead of hamburger menu icon
- When switching window in maximize mode, toolbar status is not updated

Otherwise it looks like I was thinking.
Though I'm wondering if we should show all the buttons on the bottom, and just mark active window with some visual indication such as different opacity or color. Then buttons would act like 'tab handles'. What do you think abou that? Would it be worth trying?

tuomas2:
> @tuomas2 Can you check why 1 test is failing? It seems to be textProgressionFinPR in SpeakTests.kt and I can't figure out how any of the changes in this PR affects that test. It passes fine in my Android Studio.

Tested cleaning cache and rebuilding master in travis and it fails too. (though it was ok before) So this problem is not in your code. I will try to fix that in master.

Not sure if something has changed in FinPR module.

tuomas2:
> Not sure if something has changed in FinPR module.

Fixed build by ignoring that particular test. FinPR module seem to have been broken recently. https://tracker.crosswire.org/projects/MOD/issues/MOD-351?filter=allopenissues

timbze:
> * When in maximize mode, un-maximize icon should be shown instead of hamburger menu icon

I think that's a good idea. Should we have the un-maximize button where it was before (bottom right) and the window buttons to the left of it? Or do you think it's better at the location where the hamburger icon is?

> Though I'm wondering if we should show all the buttons on the bottom, and just mark active window with some visual indication such as different opacity or color. Then buttons would act like 'tab handles'.

I think this would be a good idea for how it looks. Though it would change the way minimized windows work, a bit. I will need to look at the code some more and see what would be a good way to do that, and still have minimized windows working as they do now when not in maximized mode.

> * When switching window in maximize mode, toolbar status is not updated

Yup, I see that now. Will fix.

timbze:
> * When switching window in maximize mode, toolbar status is not updated

I have fixed the other issues now, but this one not yet. I thought I just needed to set active window and it seemed to work at first, but then it doesn't load the window in some cases when switching from one to the other and I can't figure out why yet. Will take some more time some time later.

tuomas2:
Bugs: 
 - Opening link in a 'special' window' does not show "x" button (for example strongs links).

tuomas2:
> Bugs:
> 
> * Opening link in a 'special' window' does not show "x" button (for example strongs links).

Fixed this bug and merged. Thanks!

PR #369, opened 2019-05-24
Fixed: In maximized mode, links window X button was in wrong location…

timbze:
Fixed X button in wrong location:
Screenshot_2019-05-23-05-10-06


tuomas2:
Will be fixed in buil 309.

tuomas2:
@timbze it would be better if you created ticket for testing purposes and pull request just links to the ticket.

PR #385, opened 2019-06-02
Exclude external files in automatic Google app backup

### LINKED ISSUE #365: Google app backup automatic integration
Something to look into. And Bible is coded to allow Google to make automatic backups of the entire app. But Google only does backups when the app is up to 25mb total.

It is possible to select which folders are included in the backup. It might be a good idea to have folders with preferences and DB backed up, but not modules. That way the size limit would probably never be exceeded.

timbze:
Closes #365 

This PR is required for Google app backup to work, because Google has a limit of 25mb per app. With downloaded modules, this quota is exceeded very quickly and then there are no backups to Google drive.

Google app backup saves preferences, bookmarks, notes, and reading plan status.

NOTE: User must have Backup to Google Drive enabled in Device Settings (available from Android 6.0)

### DIFF
--- a/app/src/main/res/xml/backup_descriptor.xml
+++ b/app/src/main/res/xml/backup_descriptor.xml
@@ -1,3 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <full-backup-content>
+    <exclude domain="external" path="." />
 </full-backup-content>


tuomas2:
How can this be tested / verified?

timbze:
In theory, there are 2 ways. This was working for me on Emulator with Android 9.0, but now I tried getting to work on my LG G6 Android 8.0 and it's not backing up yet. I've read something about Google backup banning the app after it gets too big. I might have to try backing up under another account, not sure. I will try it yet.
1.
Find Google Backup settings under device settings. Mine shows last backup for AndBible is from Feb 26. I must have had AB reinstalled from scratch and not added many modules yet. If AndBible is not in the list that means it probably was never small enough to back up. Google backs up approx once a day when on wifi and plugged in (I think). After applying this PR, it should back up (Not working on my LG G6 yet). To test restore, simply uninstall the app and reinstall. Either from Play Store or Android Studio works.
Screenshots

2.
Here's info on method 2, https://developer.android.com/guide/topics/data/testingbackup.html
Run `adb shell bmgr backupnow net.bible.android.activity`
If app size is too big, it will give you either below error, or I've also seen another error saying something like, "Not permitted to backup"

> timmy@HP-Z420-TB:~/Android/and-bible$ adb shell bmgr backupnow net.bible.android.activity
Running incremental backup for 1 requested packages.
Package @pm@ with result: Success
Package net.bible.android.activity with result: **Transport rejected package**
Backup finished with result: Success

timbze:
My note above about Google banning the app from backing up is false. Google's documentation https://developer.android.com/guide/topics/data/autobackup says:

> Caution: If the amount of data is over 25MB, the system calls onQuotaExceeded() and doesn't back up data to the cloud. The system periodically checks whether the amount of data later falls under the 25MB threshold and continues Auto Backup when it does.

timbze:
I finally did a factory reset on my phone, restored previous DB and all modules, and now automatic backup works. I'm guessing this won't be needed on most devices :)

PR #389, opened 2019-06-07
Bugfix AB crash after module deleted and switching tab (workspace)

### LINKED ISSUE #375: App crashes when a bible module in one of the tabs is deleted...
**Describe the bug**
AB crashes when a bible module in one of the tabs that is not in view is deleted.

**To Reproduce**
Create two tabs with two windows in each tab. first tab should have kjv and esv. For second tab use kjv,  jps
1) From the first tab, tap and hold Bible chapter, on the top left on the menu bar.
2) from the list of Bibles tap and hold JPS, then from top left of menu tap on the delete icon.
3) try to switch to second tab which use to contain jps module, the app crashes.

**Expected behavior**
Needs to be discussed.

### LINKED ISSUE #388: First Bible window is too small after adding new window
**To Reproduce**
1. Open 2 windows
2. Pull the separator bar as far down as possible
3. Close window 1 (window 2 is now the only window)
4. Add a new window

**Problem** 
The separator bar is now very high, usually behind the menu bar

**Expected behavior**
The separator bar should be centered between the 2 windows

timbze:
Closes #375 
Closes #388

### DIFF
--- a/app/src/main/java/net/bible/android/control/document/DocumentControl.kt
+++ b/app/src/main/java/net/bible/android/control/document/DocumentControl.kt
@@ -164,7 +164,6 @@ class DocumentControl @Inject constructor(
         windowControl.windowRepository.windows.filter {
             !it.pageManager.currentBible.checkCurrentDocumentStillInstalled()
         }.forEach {
-            it.pageManager.currentBible.currentDocument
             PassageChangeMediator.getInstance().onCurrentPageChanged(it)
         }
     }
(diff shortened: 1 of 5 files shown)


app/src/main/java/net/bible/android/control/document/DocumentControl.kt
```diff
@@ -157,6 +160,15 @@ class DocumentControl @Inject constructor(
         activeWindowPageManagerProvider.activeWindowPageManager.setCurrentDocument(newDocument)
     }
 
+    fun checkIfAnyPageDocumentsDeleted() {
+        windowControl.windowRepository.windows.filter {
+            !it.pageManager.currentBible.checkCurrentDocumentStillInstalled()
+        }.forEach {
+            it.pageManager.currentBible.currentDocument
```
tuomas2: isn't this line unnecessary?

timbze: Yes, you are right. Fixed

PR #419, opened 2019-07-28
Bugfix/#418 user custom reading plan name

### LINKED ISSUE #418: User custom reading plan shows as blank line
**To Reproduce**
1. Copy [this file](https://github.com/AndBible/and-bible/files/3709507/y1ot1nt1_chrono.txt) to your sdcard jsword/readingplan folder
2. Make sure And Bible is closed and not in apps list. Open And Bible and look for the reading plan in the reading plan list. (if it does not show up you may have to go under device settings > Apps > And Bible > Permissions and enable storage permission)
3. When trying to select a reading plan, you will see one line is blank in the reading plan list. You can click it and the reading plan opens correctly though.

**Expected behavior**
The reading plan list should at least show the plan file name if it has no Plan Name specified in the file.

timbze:
Closes #418

### DIFF
--- a/app/src/main/java/net/bible/service/readingplan/ReadingPlanDao.kt
+++ b/app/src/main/java/net/bible/service/readingplan/ReadingPlanDao.kt
@@ -32,6 +32,8 @@ import org.crosswire.jsword.versification.system.SystemKJV
 import org.crosswire.jsword.versification.system.SystemNRSVA
 import org.crosswire.jsword.versification.system.Versifications
 
+import java.io.ByteArrayOutputStream
+import java.io.ByteArrayInputStream
 import java.io.File
 import java.io.FileInputStream
 import java.io.IOException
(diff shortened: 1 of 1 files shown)


app/src/main/java/net/bible/service/readingplan/ReadingPlanDao.kt
```diff
@@ -196,43 +212,80 @@ class ReadingPlanDao {
...
+                if (lineCount == 0) {
+                    properties.planName = StringUtils.trim(lineWithoutCommentMarks)
+                } else {
+                    properties.planDescription = StringUtils.trim(
+                        "${properties.planDescription ?: ""} $lineWithoutCommentMarks "
+                    )
+                }
+                lineCount++
+            }
+            loopCount++
```
tuomas2: if loopCount >= 5, we could break and return?

tuomas2:
@timbze Please check / review my improvements in efb7c0de01e8728d2e271a846c9ab5d06e99e7dd

app/src/main/java/net/bible/service/readingplan/ReadingPlanDao.kt
timbze: I thought so too but seems there is no break that works for "forEachLine" method.

https://stackoverflow.com/a/48623119/877453

That answer says return@forEachLine works, but I tried all kinds of stuff (including this I think) when I wrote this and could not get them working at all.

PR #453, opened 2019-11-28
Feature/room

tuomas2:
https://developer.android.com/training/data-storage/room/

Not any tables in Room yet, but framework is in place.


app/src/main/java/net/bible/service/db/DatabaseContainer.kt
```diff
@@ -68,6 +68,25 @@ private val MIGRATION_5_6 = object : Migration(5, 6) {
     }
 }
 
+private val MIGRATION_6_7 = object : Migration(6, 7) {
+    override fun migrate(db: SupportSQLiteDatabase) {
+        db.apply {
+            execSQL("DROP TRIGGER IF EXISTS bookmark_cleanup;")
+            execSQL("DROP TRIGGER IF EXISTS label_cleanup;")
+            execSQL("CREATE TABLE `bookmark_label_new` (`bookmark_id` INTEGER NOT NULL, `label_id` INTEGER NOT NULL, PRIMARY KEY(`bookmark_id`, `label_id`), FOREIGN KEY(`bookmark_id`) REFERENCES `bookmark`(`_id`) ON UPDATE NO ACTION ON DELETE CASCADE , FOREIGN KEY(`label_id`) REFERENCES `label`(`_id`) ON UPDATE NO ACTION ON DELETE CASCADE );");
+            execSQL("CREATE TABLE `readingplan_status_new` (`_id` INTEGER, `plan_code` TEXT NOT NULL, `plan_day` INTEGER NOT NULL, `reading_status` TEXT NOT NULL, PRIMARY KEY(`_id`), FOREIGN KEY(`plan_code`) REFERENCES `readingplan`(`_id`) ON UPDATE NO ACTION ON DELETE CASCADE );");
```
timbze: Table reading_status field plan_code does not correspond to table readingplan field _id

Also, when user starts a plan, reading_status gets records added as soon as a chapter is marked as read, but readingplan only gets record added when user goes to second day. 

Both table readingplan and reading_status base on plan_code text, which is the text file name of reading plans. Table readingplan field _id is not being used at all right now actually.

tuomas2: Made fixes, you can check them out

tuomas2: also enabled foreign_keys

app/src/main/java/net/bible/service/db/DatabaseContainer.kt
```diff
@@ -71,17 +71,19 @@ private val MIGRATION_5_6 = object : Migration(5, 6) {
...
     override fun migrate(db: SupportSQLiteDatabase) {
         db.apply {
+            execSQL("PRAGMA foreign_keys=ON;")
             execSQL("DROP TRIGGER IF EXISTS bookmark_cleanup;")
             execSQL("DROP TRIGGER IF EXISTS label_cleanup;")
             execSQL("CREATE TABLE `bookmark_label_new` (`bookmark_id` INTEGER NOT NULL, `label_id` INTEGER NOT NULL, PRIMARY KEY(`bookmark_id`, `label_id`), FOREIGN KEY(`bookmark_id`) REFERENCES `bookmark`(`_id`) ON UPDATE NO ACTION ON DELETE CASCADE , FOREIGN KEY(`label_id`) REFERENCES `label`(`_id`) ON UPDATE NO ACTION ON DELETE CASCADE );");
-            execSQL("CREATE TABLE `readingplan_status_new` (`_id` INTEGER, `plan_code` TEXT NOT NULL, `plan_day` INTEGER NOT NULL, `reading_status` TEXT NOT NULL, PRIMARY KEY(`_id`), FOREIGN KEY(`plan_code`) REFERENCES `readingplan`(`_id`) ON UPDATE NO ACTION ON DELETE CASCADE );");
+            execSQL("CREATE TABLE `readingplan_status_new` (`_id` INTEGER, `plan_code` TEXT NOT NULL, `plan_day` INTEGER NOT NULL, `reading_status` TEXT NOT NULL, PRIMARY KEY(`_id`), FOREIGN KEY(`plan_code`) REFERENCES `readingplan`(`plan_code`) ON UPDATE NO ACTION ON DELETE CASCADE )")
```
timbze: I am not sure if it is a good idea to have a foreign key with plan_code. Will child table be able to have a record without parent table having that plan_code?

As I said in my other comment, when user starts a plan, reading_status gets records added as soon as a chapter is marked as read, but readingplan only gets record added when user goes to second day. I'm not sure if that will be allowed with the foreign key?

tuomas2: Okay, dropped foreign key from here.

timbze: I think you missed line 78?

tuomas2: True, fixed

Issue #452, opened 2019-11-28
Bookmark and My Notes code refactored to Kotlin


tuomas2:
Testing needed.

- Database handling operations, like inserting, updating, getting, etc.
- Backup & restore

Will be in build 320.0.

tuomas2:
I guess these are tested by now -> closing.

PR #462, opened 2019-12-09
Bugfix/#459 Fix note to open at correct verse

### LINKED ISSUE #459: My note window opens at wrong verse
Happens only after 319.4, after Kotlinizing MyNotes I think.

**To Reproduce**
1. Scroll to Matt 1:10, so v10 is the top verse in screen
2. Select verse 12 and click the pencil icon to open MyNotes
3. The note opens on v10, but it should be the selected verse, v12

timbze:
I don't know if this is the best fix but it works
Closes #459 
Testing in progress #452


tuomas2:
I think this is a bit hacky, but I'll accept it this time.

PR #561, opened 2020-02-25
Feature/pin mode

tuomas2:
I'm experimenting a new, improved window behavior settings.

- Remove maximize mode
- Add pin flag to windows 
- Auto-pin option = true (from main options menu) makes it work like previous non-maximized mode
- Auto-pin option = false with no pinned windows makes it work like previous maximized mode


app/src/main/java/net/bible/service/db/DatabaseContainer.kt
```diff
@@ -258,7 +258,7 @@ private val MIGRATION_18_19 = object : Migration(18, 19) {
 private val MIGRATION_19_20 = object : Migration(19, 20) {
     override fun migrate(db: SupportSQLiteDatabase) {
         db.apply {
-            val colDefs = "`isSwapMode` INTEGER NOT NULL DEFAULT 0".split(",")
+            val colDefs = "`isSwapMode` INTEGER NOT NULL DEFAULT FALSE".split(",")
```
tuomas2:
```suggestion
            val colDefs = "`isSwapMode` INTEGER NOT NULL DEFAULT 0".split(",")
```

Issue #630, opened 2020-04-26
Search view is opened twice


JJK96:
If a match appears a lot of times in the document, the search can take a lot of time.

For me, a search for "en" (Dutch for "and") in a Dutch translation yields more than 1000 results and takes about 23 seconds on my device.

My device is a OnePlus 6 with Android 10.

tuomas2:
has it been faster earlier?

JJK96:
I don't think so, but I did not test with a recent version.
Is it faster for you with a 1000+ result search?

tuomas2:
Slow here too. It should not block the Ul, but it does (Should be fixed with kotlin Coroutines). but probably not a regression.

JJK96:
When debugging, I discovered, that the search is actually done twice before showing the result.

The `SearchResults::onCreate()` is called twice.

This is probably also what causes the window to disappear and reappear.

JJK96:
Here is a video about the find windows opening twice.

https://youtu.be/2MFwpVR_ubo

The first time `SearchResults::onCreate()` is called, it is with `savedInstanceState == null` and the second time with an initialized `savedInstanceState`.

JJK96:
The cause of this issue is that the activity is restarted due to a theme change, I can reproduce the issue if I have system dark theme on in android 10.
If I disable the dark theme, the issue disappears. 

Possible fixes are:

- add `uiMode` to the `configChanges` field of the `Search` and `SearchResults` fields in `AndroidManifest.xml`.
     + This is confirmed to work, but I think it is not the right way to fix it.
- Actually handle the night theme correctly by making the theme extend `Theme.AppCompat.DayNight`.
    + I am experimenting with the this solution, but it will probably result in a complete overhaul of the theming nightmode implementation.

tuomas2:
Changed code so that it shows progress indicator when loading results and does not block UI 6a575916b6a741146eb829cb8c59c4fdd2925ae1

PR #634, opened 2020-04-30
Improve search speed by using an iterator instead of .get()

JJK96:
Ref #630


JJK96:
I see the spacing is off, let me fix that.

PR #637, opened 2020-04-30
Make NoteRef and MyNoteMarker clickable

JJK96:
Click on the MyNotes image or NoteRef to go to the corresponding activity

It works by wrapping the reference text in `<a> </a>` with a special url format, which is parsed by the UriAnalyzer to open the right activity

PR #649, opened 2020-05-09
Fix tests that failed due to merge of #637

JJK96:
Also added space after note like there used to be.

### DIFF
--- a/app/src/test/java/net/bible/service/format/osistohtml/taghandler/NoteHandlerTest.java
+++ b/app/src/test/java/net/bible/service/format/osistohtml/taghandler/NoteHandlerTest.java
@@ -1,5 +1,6 @@
 package net.bible.service.format.osistohtml.taghandler;
 
+import net.bible.service.common.Constants;
 import net.bible.service.format.Note;
 import net.bible.service.format.osistohtml.HtmlTextWriter;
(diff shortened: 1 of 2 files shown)


app/src/test/java/net/bible/service/format/osistohtml/taghandler/NoteHandlerTest.java
```diff
@@ -71,7 +82,7 @@ public void simpleNote() {
 		noteHandler.end();
 		writer.write("after note");
 		
-		assertThat(writer.getHtml(), equalTo("before note<span class='noteRef'>a</span> after note"));
+		assertThat(writer.getHtml(), equalTo(String.format("before note<a href='%s:%s%d/a' class='noteRef'>a</a> after note", Constants.NOTE_PROTOCOL, basisRef.getOsisID(), verseInfo.currentVerseNo)));
```
tuomas2, line 85: In tests, you shouldn't do like that. Let's say, you accidentally change Constants.NOTE_PROTOCOL. The test should fail, but it does not. Tests should be broken (and fixed) if such changes are made, accidentally or not.

JJK96: I disagree, if the note protocol constant is changed, the functionality will still be the same, so the test should not fail.

PR #653, opened 2020-05-09
Bugfix clickable notes first chapter

JJK96:
Get the current verse from the verseInfo, instead of basisRef. This is more reliable.

Also update the tests to match and added a regression test.


tuomas2:
Looks good. Thanks!

PR #663, opened 2020-05-21
Merge AppThemeDay and AppThemeNight into one DayNight theme

### LINKED ISSUE #621: System night mode switching broken
Night mode: "System" does not work (Android 10 feature, though some Samsung Android 9 also impelment it).

### LINKED ISSUE #630: Search view is opened twice

JJK96:
Please test this thoroughly on your devices before merging.

Also ping me if you want to merge so I can squash the commits, these commit messages are not production ready.

Fixes #621 and fixes #630


JJK96:
Speak view, `-stopped-` text is white on white.

tuomas2:
If this requires a lot of testing and has possibility of breaking things greatly, I want to postpone this to past-release time.

JJK96:
That sounds like a good idea

tuomas2:
Testing on emulator running Android 10.
Bugs:
 - [ ] In day mode, navigation bar color wrong on other than main activity

JJK96:
The color of the navigation bar is different from the old theme, but do you mean that the color changes when you open another activity? 
Do you have screenshots, since for me the navigation bar color is the same for every activity.

tuomas2:
Here are screenshots:
Should look like this:
Screenshot_1592476989

Looks like this:
Screenshot_1592476863

You can try yourself in emultator (Android 10, API 29) too (old code vs new code).

JJK96:
Your screenshots show the same color that I see. 
I removed some of the custom colors and found that the default android colors looked fine (to me), so this would simplify the theme a bit.

Do you disagree?
Do you want the colors to be exactly the same as the current theme?

timbze:
Android 8.0 on LG G6 does not show everything correctly in night mode. See pics:
Edit: I removed most pics now because I longer have that issue. Not sure why I had them before.

This issue remains: reference color in fullscreen here has gotten an odd color:
image

timbze:
I think I have now fixed the rest of issues mentioned in the comments.

0df6fc9 is about a separator line in dark theme that looks bad (or distracting) to me (also was mentioned in AB Support group on June 11)

tuomas2:
Okay, let's continue testing in develop :-)

PR #708, opened 2020-06-07
Add stetho tool to be used for DEBUG build only

timbze:
This is very helpful in checking db (and other stuff) while debugging etc. Check more info here: http://facebook.github.io/stetho/

Your call @tuomas2 if you want to add this.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -170,7 +170,7 @@ dependencies {
     implementation 'org.apache.lucene:lucene-analyzers:3.6.2'
     implementation 'de.psdev.slf4j-android-logger:slf4j-android-logger:1.0.5'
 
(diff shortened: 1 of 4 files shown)


tuomas2:
It can be enabled, though I need to get into what it is a little more before merging. We need to re-check that it does not increase apk size (I think proguard should make sure it does not).

timbze:
> We need to re-check that it does not increase apk size (I think proguard should make sure it does not).

It did increase Release apk by 150kb. I have now pushed a change that keeps Release apk exact same size

tuomas2:
Thank you for introducit this Stetho tool to me. It's indeed cool and helps in debugging.

Issue #688, opened 2020-06-01
Make Travis build APK for testers


tuomas2:
It could be nice and beneficial if Travis could build and upload somewhere APKs of successfull pull request builds, for beta testers.

Maybe not yet worth it but if we get a lot of pull requests, it could be helpful.

timbze:
Kind of separate idea, I know some platforms have a separate branch that they set up to auto-build. So for example, when ready to release beta, merge to beta branch, and there can be pipelines set up to build apk and add to beta tag under releases.

But this would be different from building together with PRs.

tuomas2:
Actually it would be possible to experiment with Github Actions too as CI replacement.

CI could be employed more for various tasks, but I'm not sure if it's worth the trouble yet. But of course, pull requests (for configuring something more fancy) are welcome :-)

timbze:
@tuomas2 Github does not allow apk attachments, but zipped yes. But I guess that would be fine?

tuomas2:
Less inconvenient than clicking apk directly, but I guess better than nothing. We can try.

tuomas2:
It is possible to upload apk somewhere (repository for example) and put a link to it.

timbze:
> It is possible to upload apk somewhere (repository for example) and put a link to it.

I think we may need to do that. It seems that it's impossible to attach files to pr or issue comment programmatically. I contacted github about it to see if I'm missing something. 

Edit: Github confirmed my thoughts and said this:
> There's currently no API for comment file attachments at the moment unfortunately so it's not possible to do this programatically. Sorry I don't have better news but if you have questions, please let me know.

[See this issue](https://github.com/isaacs/github/issues/1133) in a Github staff repo

timbze:
Closing in favor of #730

PR #727, opened 2020-06-15
Improvement - Add GitHub actions CI

timbze:
This would make Github actions do what TravisCI is currently doing.

1. This would also require adding repo secret named TEST_MODULE_KEY with the openssl decryption key.
2. Also requires adding secret named DOWNLOAD_TEST_MODULES_PATH with path to encrypted modules file.

You can check how this works, here: https://github.com/timbze/and-bible/actions

Encrypt the modules on linux like this: 
`openssl enc -in testmods.zip -out testmods.zip.enc -e -aes-256-cbc -pbkdf2 -pass 'pass:TEST_MODULE_KEY'`

**Possible negative side effects?**
~~testmods.zip.enc file is 21mb, which all developers would have an additional 21mb storage in use. Possible change would be to add the file on andbible html repo (or another one) and download from there.~~ Updated yml to download from specified http(s) directory.

**Part 2 - Create release for tags**
Whenever a tag is pushed that beings with build- , it will auto build and sign apk, create beta release and attach the apk.

_This could be done in 2 ways:_
**Current way:** With "if" conditions for the steps that are about release, to make sure if the push is a tag
**Another possible way:** Add another job, for example, we have "build" job now. We could put the first job as "test", and add a second job "release" that would only run if push is a tag AND first job succeeded. However, each job has to set up everything, so that means quite a bit of the code would have to duplicate on job "release"

Requirements
Repo secrets:
JKS_BASE64 (create text with openssl base64 -A -in keystorefile.jks)
KEYSTORE_PASS
KEY_PASS

Related: #688 (maybe I can continue enhancement once this is up and running well)

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -24,19 +24,24 @@ jobs:
         # Version Spec of the version to use.  Examples: 10.x, 10.15.1, >=10.15.0
         node-version: 10.x
 
-    - name: Cache gradle dependencies, test modules etc
-      id: cache
+    - name: Cache gradle dependencies
       uses: actions/cache@v2
       with:
         path: |
           ~/.gradle/caches
           ~/.gradle/wrapper
           ~/.android/build-cache
-          ~/.sword
-        key: ${{ runner.os }}-cache-20200615 # cache auto-removes if not accessed in 7 days. If data needs to refresh, change key name here (till https://github.com/actions/cache/issues/2 is addressed)
+        key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
+
+    - name: Cache test modules
+      id: cache-test-modules
+      uses: actions/cache@v2
+      with:
+        path: ~/.sword
+        key: ${{ runner.os }}-test-modules # cache auto-removes if not accessed in 7 days. If data needs to refresh, change key name here (till https://github.com/actions/cache/issues/2 is addressed)
 
     - name: Download test modules if not cached
-      if: steps.cache.outputs.cache-hit != 'true'
+      if: steps.cache-test-modules.outputs.cache-hit != 'true'
       run: |
         wget ${{ secrets.DOWNLOAD_TEST_MODULES_PATH }} -O ./testmods.zip.enc
         openssl enc -in ./testmods.zip.enc -out ./testmods.zip -d -aes-256-cbc -pbkdf2 -pass 'pass:${{ secrets.TEST_MODULE_KEY }}'
@@ -46,7 +51,7 @@ jobs:
     - name: Build with Gradle
       run: ./gradlew --console plain build
 
-    - name: Before cache
+    - name: Before saving cache
       run: |
         rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
         rm -fr $HOME/.gradle/caches/*/plugin-resolution/
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -27,14 +27,23 @@ jobs:
     - run: mkdir testModules
 
     - name: Get modules from cache if exists
-      id: cache
+      id: cache-modules
       uses: actions/cache@v2
       with:
         path: testModules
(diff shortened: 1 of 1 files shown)


tuomas2:
Yes I think it's better to download it from somewhere else rather than add to repository.

tuomas2:
I can add it to https://github.com/AndBible/data for example (this is served under andbible.github.io)

tuomas2:
Can you compare speed of the builds? 
Could we use caching like in travis?

timbze:
**Speed**
4 runs were between 11.5 and 12.5 minutes
TravisCI last some tests were 13-14 minutes.

**Cache**
I didn't check into caching yet since the file was local like this. But we can cache, updating PR now.

tuomas2:
There are more things to cache than just modules zip file. We have been caching gradle dependencies etc in .travis.yml.

timbze:
> There are more things to cache than just modules zip file. We have been caching gradle dependencies etc in .travis.yml.

Now caching same stuff as TravisCI (as far as I know)

One think lacking about this cache feature is https://github.com/actions/cache/issues/2. When test-modules would change, or gradle version changes, there is no way to clear cache but to change cache key (or if cache expires when not used for 7 days). Also noted in yml file. 

When gradle version changes there is no need to clear cache, just that the size of cache will keep growing. Right now cache is ~665mb and we can store up to 5GB cache, so shouldn't be an issue on that part.

timbze:
Runs in <10 minutes after caching all those files

tuomas2:
Growing cache size probably will slow down the build a bit, because cache is fetched always before build starts.

timbze:
I think it would grow only if gradle version changes, cause then both
version wrappers are stored. Which I assume Travis would do the same
without a manual cache clear.

timbze:
Actually, this cache never get's updated, only replaced. So the newer gradle version would not get cached at all without either:

1. A manual change of the key or
2. Using hashFiles as displayed in README [here](https://github.com/actions/cache). Like this (though the files to hash would probably need to be something else): `key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}`

timbze:
Have now changed gradle cache to use hashFiles in key as example [here](https://github.com/actions/cache/blob/master/examples.md#java---gradle)

.github/workflows/android.yml
```diff
@@ -0,0 +1,57 @@
+name: Android CI
```
tuomas2, line 1: Name could be "Android tests"

tuomas2: name of the file could be accordingly tests.xml

tuomas2: okay not any more.

timbze:
@tuomas2 as you mentioned in #728, we could use different tag prefixes for different things... So we should also plan for that.

Something like beta- and release- ? Or for release do as we did with v3.3.380 etc ?

Or we could continue the same build- for beta and v3.3.380 for releases... But we can also change to something else if we'd want.

tuomas2:
Hmm. maybe beta-X for beta releases on play store. build-X would also build APK and release only on github. production-X for releases. I will use vX.Y.Z tag for major releases too, but I like that automatic CI action tags are bit different.

timbze:
I have now updated PR to build github release apk on build-, beta-, and production- tags

I would love setting up fastlane yet for that part, but I won't be able to spend more time on this now.

**One other note**, the cache has issues some of the time yet, but it is being addressed [here](https://github.com/actions/cache/issues/267). The only issue is that occasionally it cannot restore, so then it takes long to try to restore, which doesn't work, and then it has to download all the data again and that takes longer too. So it's just a longer run on some occasions.

tuomas2:
> I have now updated PR to build github release apk on build-, beta-, and production- tags
> 
> I would love setting up fastlane yet for that part, but I won't be able to spend more time on this now.

That's good start already. Let's see if I will try Fastlane from this basis or if it will be left for future enhancement.

PR #728, opened 2020-06-16
Github actions release apk on build- tag

timbze:
Whenever a tag is pushed that beings with build- , it will auto build and sign apk, create beta release and attach the apk.

**Requirements**
Repo secrets: 
JKS_BASE64 (create text with `openssl base64 -A -in keystorefile.jks`)
KEYSTORE_PASS
KEY_PASS

### DIFF
--- a/.github/workflows/android-release.yml
+++ b/.github/workflows/android-release.yml
@@ -1,0 +1,77 @@
+name: Release apk
+
+on:
+  push:
+    tags: [ build-* ]
+
+jobs:
+  buildApk:
+
+    runs-on: ubuntu-20.04
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up JDK
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+
+    - name: Set up Node.js environment
+      uses: actions/setup-node@v2-beta # v2-beta adds caching
+      with:
+        # Version Spec of the version to use.  Examples: 10.x, 10.15.1, >=10.15.0
+        node-version: 10.x
+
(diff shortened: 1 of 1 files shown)


tuomas2:
does apk go to github releases page?

ildar:
line 42 and below: you should decide for yourself. This is fairly optional.

tuomas2:
> line 42 and below: you should decide for yourself. This is fairly optional.

you  mean signing?

ildar:
signing and further (uploading, etc.)

tuomas2:
@timbze maybe next step would be use `fastlane supply` to release also on beta channel (perhaps with another tag prefix). And perhaps even to production with yet another tag...

tuomas2:
In general this will save a lot of time from me in the long run (not having to upload manually to drive (and play store beta channel, later)). So thanks a lot! :-) 

Maybe we could make this conditional such that it does not release if tests fail (so chaining with another action somehow?).

timbze:
> Maybe we could make this conditional such that it does not release if tests fail (so chaining with another action somehow?).

I would need to add it to my first PR. 

We can do it 2 ways:
1. With "if" conditions for the steps that are about release, to make sure if the push is a tag (added to #727 like this for now)
2. Add another job, for example, we have "build" job in the other file now. We could put the first job as "test", and add a second job "release" that would only run if push is a tag AND first job succeeded. However, each job has to set up everything, so that means quite a bit of the code would have to duplicate on job "release"

timbze:
I think we should probably do it in only 1 file anyway for now. Closing this.

Issue #636, opened 2020-04-30
Backup to phone storage does not work on android 10


JJK96:
**Describe the bug**
When saving the application database to the sdcard "an error occurs".

**Bug was found on And Bible version**
2e5a3e5b9fee911498faba95623c5a170d17c48d

**To Reproduce**
Go to Administration > Backup & Restore > Backup application database to... > Phone storage

**Expected behavior**
File is saved to SD card

**Smartphone (please complete the following information):**
 - Device: Oneplus 6
 - OS: Android
 - Version 10.3.3

**Additional context**
While debugging I found that the problem is that `toDir.mkdir()` returns false, indicating that the directory could not be created. This results in an error on the directory path, because it does not exist. (`java.lang.IllegalArgumentException: Invalid path: /storage/emulated/0/andbible_backup`)

According to [this QA](https://stackoverflow.com/questions/56821095/android-q-file-mkdirs-returns-false) writing to an arbitrary location on external storage is no longer possible on Android 10.

I think the best course of action is to remove the save to phone storage option since the share option can be used to make another application do the work, plus the user knows in which location it is saved.

timbze:
I think we need to use something like Intent(Intent.ACTION_GET_CONTENT). This allows user to pick location to save file and I assume that will still work in Android 10.

This "backup to phone" feature was removed some time ago but recently put back in, see #600

bvahedy:
I like to keep this feature as it requires no internet connection to save
your database and it also takes less time to restore the database back.

tuomas2:
If it's causing error in android 10 then it will be disabled for android 10+.

But if someone would like to investigate better options to "share" that allows user to select folder from local storage, it would be best. I know there is / are some app(s) that can do it as "share" target, but is there anything better?

tuomas2:
@JJK96 More informative would be if there would be something like "Bug was found on And Bible version 2e5a3e5 (366+)" instead of plain git hash.

tuomas2:
@JJK96 Tested with Android 10 emulator and it works. QA says that it does not work any more with Android 11.

JJK96:
I just tested again, for me it does not work, are you sure that you had not already created the `andbible_backup` directory in the emulator?

tuomas2:
Tried again, removed folder and re-created backup and it looks fine.

JJK96:
After creating the folder manually on my device:

```
E/net.bible.service.common.FileManager: Error moving file
[...]

PR #740, opened 2020-06-25
Bug fixes for backing up on Android 11

### LINKED ISSUE #739: Backing up modules or application db via intent crashes in Android 11
When a user goes to select the backup option, and then back up either the database or the modules, the 
app immediately crashes

Version 3.3.381

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to menu
2. Click on Backup & Restore
3. Click `Backup Documents to...`
4. Select 1 or more modules
5 .Press OK
6. Crash

**Expected behaviour**
No crash

**Smartphone (please complete the following information):**
 - Device: Pixel 4 XL
 - OS: Android 11 
 - Version  11 (RPB1.200504.020)

**Additional context**
.

mattschlosser:
I was trying out this app and found that it crashed in Android 11 when trying to backup the modules or app db when using the share button (i.e. via Intent). I figured I'd take a stab at trying to fix this bug.

This is similar to #636, but does not fix that bug, as I did not use the phone storage option.

I don't have much experience working with Android. Any feedback is appreciated. 👍 

The root of the problem seems to be that Android 11 changes the internal path structure of Android apps.
The internal path structure now looks like `/data/user/0/<PACKAGE_NAME>` instead of `/data/data/<PACKAGE_NAME>`
The solution I came up with was to grab the app's storage location from the app's Context, passing it into BackupControl in the onCreate function when the app is started.  Using this, I was able to 
briefly test out the fixes on 3 devices, one on Android 7, one on Android 10, and one on Android 11, and all 3 seemed to work fine after the changes I made. 

Closes #739

### DIFF
--- a/app/src/main/java/net/bible/android/control/backup/BackupControl.kt
+++ b/app/src/main/java/net/bible/android/control/backup/BackupControl.kt
@@ -303,8 +303,13 @@ class BackupControl @Inject constructor() {
     companion object {
 
         // this is now unused because And Bible databases are held on the SD card to facilitate easier backup by file copy
-        private val internalDbDir = File(Environment.getDataDirectory(), "/data/" + SharedConstants.PACKAGE_NAME + "/databases/")
-        private val internalDbBackupDir = File(Environment.getDataDirectory(), "/data/" + SharedConstants.PACKAGE_NAME + "/files/backup")
+        private lateinit var internalDbDir : File;
+        private lateinit var internalDbBackupDir: File;
+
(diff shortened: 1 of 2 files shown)


tuomas2:
Thank you for fixing this issue! Looks good to me. Could you try to fix #636 similar way? 
I can merge this to master and cherry-pick to release/v3.3, but otherwise I would recommend making bugfixes towards release/v3.3 branch for now.

If you might consider contributing more in the future, please send me email! My email is tuomas.airaksinen@gmail.com.

PR #748, opened 2020-06-27
Fixes #636 - Allows user to pick backup location on device

### LINKED ISSUE #636: Backup to phone storage does not work on android 10

mattschlosser:
# Description

Hello again, 

Closes #636 

On newer versions of Android, the db backup to and restore from phone storage does not work. This is because of ways
shared file storage is changing in current/future versions of Android. 

Again, I'm a bit of an Android newbie, so feedback is appreciated 👍 

### Backing up
To fix this, I used the Intent `ACTION_CREATE_DOCUMENT` which allows the user to pick where they want to save the file on their device. After choosing, the app is given temporary permission by Android to write to that location. Because access to the entire
phone storage is no longer necessary to just write one backup file, it allows us to remove the WRITE_EXTERNAL_STORAGE option
as I could not find it being used anywhere else.

### Restoring
Since there was a file picker already implemented, the user can just restore the database by clicking the Pick File option from the menu that used to come up, and because the restore from phone storage button was broken on Android 10+, I think it was best
just to remove that menu and show the file chooser directly after clicking the backup database menu option. Users can then
navigate to where they chose to backup to, or, on older versions, to the `/andbible_backup/andBibleDatabase.db` on their internal storage. 

# Possible negative side effects? 
Users may have difficulty locating the backup file if they are looking for the backup file from an older version of And Bible. The file picker hides the internal storage by default, and just shows things like "Downloads", "Audio", and "Images". Users have to click the menu in the file picker and choose 'Show internal storage"

# **Screenshots**
## Backing up
Screenshot_20200626-201237
*User clicks the phone storage option*

Screenshot_20200626-201241
__User can pick where they want to save the file. The default filename is automatically suggested, and is the same as `DATABASE_NAME`__

Screenshot_20200626-205012
*The user sees a confirmation message*

Perhaps it is no longer necessary to display the location, as this path doesn't mean anything to the users, and the user knows where they saved the file anyway.

## Restoring
Screenshot_20200626-205035
*The user selects the restore backup option*

Screenshot_20200626-205040
*They immediately see the file picker, which used to be an option on the menu that would pop up*

### DIFF
--- a/app/src/main/java/net/bible/android/control/backup/BackupControl.kt
+++ b/app/src/main/java/net/bible/android/control/backup/BackupControl.kt
@@ -53,7 +53,12 @@ import org.crosswire.jsword.book.Book
 import org.crosswire.jsword.book.BookCategory
 import org.crosswire.jsword.book.Books
 import org.crosswire.jsword.book.sword.SwordBookMetaData
-import java.io.*
+import java.io.BufferedInputStream
+import java.io.File
+import java.io.FileInputStream
+import java.io.FileOutputStream
+import java.io.IOException
+import java.io.InputStream
 import java.util.zip.ZipEntry
 import java.util.zip.ZipOutputStream
 import javax.inject.Inject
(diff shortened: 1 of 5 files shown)


tuomas2:
Looks good to me. The only thing I'm considering is if I want functional change still to 3.3.
It would be possible to make this as a fix for Android 10/11 users such that for earlier versions
it would work as earlier.

Second option is to merge is to target this to 3.4 release instead of 3.3-bugfix.

app/src/main/java/net/bible/android/control/backup/BackupControl.kt
```diff
@@ -52,11 +53,7 @@ import org.crosswire.jsword.book.Book
 import org.crosswire.jsword.book.BookCategory
 import org.crosswire.jsword.book.Books
 import org.crosswire.jsword.book.sword.SwordBookMetaData
-import java.io.BufferedInputStream
-import java.io.File
-import java.io.FileInputStream
-import java.io.FileOutputStream
-import java.io.InputStream
+import java.io.*
```
tuomas2, line 56: Just personal preference coming from my python background, but I don't really like importing *. 
I know IDE likes to do this automatically by default, but it's configurable in IDE settings.

timbze:
I think this is a good improvement but my opinion is also that if fixing for 3.3 then it should only be for Android 11 and/or 10

mattschlosser:
I made some changes so that this fix only applies to Android 10 and above. Users should see no change if they are on Android 9 or below.

app/src/main/java/net/bible/android/control/backup/BackupControl.kt
```diff
@@ -93,6 +95,39 @@ class BackupControl @Inject constructor() {
...
+     */
+    fun backupDatabaseToUri( uri: Uri) : Boolean {
+        val out = BibleApplication.application.contentResolver.openOutputStream(uri)!!
+        val filename = DATABASE_NAME;
+        val f = File(internalDbDir, filename);
+        val inputStream = FileInputStream(f)
+        val header = ByteArray(16)
+        inputStream.read(header)
+        var ok = false
+        if (String(header) == "SQLite format 3\u0000") {
```
tuomas2, line 108: Why are we checking format of a known db file we are writing? Checking is needed when restoring db, but not so much when writing, I think.

PR #752, opened 2020-06-28
Reading plan current day bugfix

### LINKED ISSUE #751: Reading plan "current day" gets reset
Build 382

**To Reproduce**
1. Start from new And Bible install
2. Open a reading plan and set to another day (e.g. go to Day 4 > Menu > Done > Click Done button, -> Reading plan window closes and day 5 has been set as current day)
3. Open reading plans again and it will open at day 5
4. Select and open another reading plan
5. Come back to the first reading plan you opened
6. It now opens on Day 1, but should have opened day 5

timbze:
Fixes #751

### DIFF
--- a/db/src/main/java/net/bible/android/database/AppDatabase.kt
+++ b/db/src/main/java/net/bible/android/database/AppDatabase.kt
@@ -26,7 +26,7 @@ import net.bible.android.database.readingplan.ReadingPlanEntities
 
 import java.util.*
 
-const val DATABASE_VERSION = 32
+const val DATABASE_VERSION = 33
 
 class Converters {
     @TypeConverter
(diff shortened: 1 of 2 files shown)


db/src/main/java/net/bible/android/database/AppDatabase.kt
```diff
@@ -26,7 +26,7 @@ import net.bible.android.database.readingplan.ReadingPlanEntities
 
 import java.util.*
 
-const val DATABASE_VERSION = 32
```
tuomas2, line 29: Version incremented but no migration?

timbze: The migration is already there. I can find later which commit it should have been incremented, but I think this was missed.

tuomas2: It looks like it has been forgotten indeed. Then it's okay to fix that.

Issue #730, opened 2020-06-18
CI improvements


tuomas2:
When everything seem to be working fine with Actions
- [x] remove .travis.yml
- [x] Update README (there's travis status banner, can we get similar for actions?)
- [ ] Test reporting, upload report html to some repo to see? 
- [ ] Try to get fastlane working 
- [ ] for pull requests, produce APK and post it on the pull request comments (see discussion on #688)
- [x] for production and beta releases, description for apk release could be taken from fastlane changelog txt automatically
- [x] title for beta / production releases should be "Release 3.3.381" or "Release 3.3.381-beta"
- [ ] It might be cool if, in beta releases, release notes would havelist of commit messages (truncated, if long), or even better, list of issues closed.
- [ ] Add default label and milestone for issues created with fire: Label = Bug and milestone = (now 3.3-bugfix). Maybe have a properties file in repo where default label(s) and milestone can be changed?
- [x] Merge current-stable to (release/{current} and develop) branch automatically (if no conflicts)

tuomas2:
About idea about posting APKs per pull requests. Although it would be convenient for testing purposes, but we should not create APKs with our official signature for any random contributed pull requests. 

I'm not sure about specifics how APK signing works, but if it would be possible to create signature that expires rather soon, then it might be ok.

timbze:
> About idea about posting APKs per pull requests. Although it would be convenient for testing purposes, but we should not create APKs with our official signature for any random contributed pull requests.

Do you mean a separate APK for each PR? Or with all PRs together? I'm trying to think how that would work. If I'm one of the people who is interested in testing 2 different PRs, that would still be impossible.

One idea might be to (could all be done automatically):
1. Change the id of the APK for PRs, 
2. Change the app title for home screen, that it would be impossible to mistake for the real And Bible.
3. And set up a different signature for that APK (might not be necessary since it's a different id but probably be better)

timbze:
> * for production and beta releases, description for apk release could be taken from fastlane changelog txt automatically
> * title for beta / production releases should be "Release 3.3.381" or "Release 3.3.381-beta"

I think we could (and should) work this out.

FYI, we can also set CI to create release as draft, so that it could be manually edited before publish. I got an email for "Release production-381", but saw @tuomas2 had manually edited it afterward. Instead of doing that, it might be better to set Release to draft in the meanwhile till we get it set up automatic.

tuomas2:
I think we can't merge all PRs together automatically. 

[...]

PR #765, opened 2020-07-07
Improve CI release text and add release notes

timbze:
Addresses #730:
- for production and beta releases, description for apk release could be taken from fastlane changelog txt automatically
- title for beta / production releases should be "Release 3.3.381" or "Release 3.3.381-beta"

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -53,11 +53,14 @@ jobs:
 
     # Test complete, create apk etc
 
-    - name: Get tag name
+    - name: Get variables
       id: tag_name
       run: |
         echo ::set-output name=TAG::${GITHUB_REF#refs/tags/}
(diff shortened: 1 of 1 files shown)


tuomas2:
What if changelog txt does not exist? I think it would be better it would not crash then either.

tuomas2:
It might be cool if, in beta releases, there would be list of commit messages (truncated, if long), or even better, list of issues closed.

timbze:
> What if changelog txt does not exist? I think it would be better it would not crash then either.

It does not crash.
This runs first (creates empty file). And if the file is empty we get the same result as now (i.e. just showing latest commit msg)
`cat > ./RELEASENOTES.txt`

If the changelog file exists, the empty file is replaced with the changelog file and the Github release body gets that text

timbze:
> What if changelog txt does not exist? I think it would be better it would not crash then either.

Changed code now to only run `cp` when the file exists

PR #768, opened 2020-07-08
Improve CI release text and add release notes

timbze:
Forgot that I should have made PR #765 toward 3.3 branch


tuomas2:
I think these configs are read (only) from default branch (here master), so this should not matter actually.

Issue #756, opened 2020-07-02
Backup improvements (for 3.4)


tuomas2:
- [ ] Use new backup system for android versions
- [ ] Use new backup system for backuping documents too
- [ ] When saving backup, location is not useful. (due to #748)

![Screenshot_1593688375](https://user-images.githubusercontent.com/5811789/86352576-ab06de80-bc6e-11ea-970f-9cef662a6785.png)

![Screenshot_1593688434](https://user-images.githubusercontent.com/5811789/86352573-a9d5b180-bc6e-11ea-9730-27b32a7a6e6b.png)

tuomas2:
@mattschlosser

tuomas2:
These changes can be towards master branch.

tuomas2:
Will be implemented in develop branch build 390.

PR #775, opened 2020-07-11
Fixes issue so users on Android 10+ can backup to locations other than just phone storage

mattschlosser:
A small change... Just moved an if statement... 😄 

This change allows users on Android 10+ devices to backup database to locations other than just their phone storage.

Related #756 

I think for 3.4, I would add the same two options when backing up modules, so once users have chosen which modules to back up, they would see two options: 
 - save to phone storage (use system file picker to save file offline)
 - share (which is currently the default/only option) (used for online sharing)

…

**Screenshots**


tuomas2:
Your plan for 3.4 sounds good

PR #776, opened 2020-07-11
Implements Backup Improvements for 3.4

### LINKED ISSUE #756: Backup improvements (for 3.4)

RuggMatt:
- [x] Users on all Android versions are now asked where they want to save the file when picking the phone storage option from the backup system
- [x] When backing up modules, users now have an option to choose phone storage
- [x] When a backup is completed, the location no longer shows
- [x] Removed unnecessary WRITE_EXTERNAL_STORAGE permissions since it is no longer being used to directly write to external storage.

Closes #756 

**Screenshots**

### DIFF
--- a/app/src/main/java/net/bible/android/control/backup/BackupControl.kt
+++ b/app/src/main/java/net/bible/android/control/backup/BackupControl.kt
@@ -41,6 +41,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 import net.bible.android.activity.BuildConfig
 import net.bible.android.database.DATABASE_VERSION
+import net.bible.android.view.activity.page.MainBibleActivity
 import net.bible.android.view.activity.page.MainBibleActivity.Companion.mainBibleActivity
 import net.bible.android.view.util.Hourglass
 import net.bible.service.common.CommonUtils
@@ -113,7 +114,7 @@ class BackupControl @Inject constructor() {
 
         if (ok) {
             Log.d(TAG, "Copied database to chosen backup location successfully")
-            Dialogs.instance.showMsg(R.string.backup_success, uri.path)
+            Dialogs.instance.showMsg(R.string.backup_success)
         } else {
             Log.e(TAG, "Error copying database to chosen location.")
             Dialogs.instance.showErrorMsg(R.string.error_occurred)
@@ -150,7 +151,7 @@ class BackupControl @Inject constructor() {
         f.delete()
     }
 
-    /** backup database from custom source
+    /** restore database from custom source
      */
     fun restoreDatabaseViaIntent(inputStream: InputStream): Boolean {
         val fileName = DATABASE_NAME
@@ -279,8 +280,33 @@ class BackupControl @Inject constructor() {
         internalDbBackupDir.deleteRecursively()
     }
 
-    suspend fun backupModulesViaIntent(callingActivity: Activity) = withContext(Dispatchers.Main) {
-        val fileName = "modules.zip"
+    fun backupModulesToUri(uri: Uri) : Boolean {
+        // at this point the zip file has already been created
+        val fileName = MODULE_BACKUP_NAME
+        val zipFile = File(internalDbBackupDir, fileName)
+        val out = BibleApplication.application.contentResolver.openOutputStream(uri)!!
+        val inputStream = FileInputStream(zipFile)
+        var ok = true
+        try {
+            out.write(inputStream.readBytes())
+            out.close()
+        } catch (ex: IOException) {
+            ok = false
+        }
+
+        if (ok) {
+            Log.d(TAG, "Copied modules to chosen backup location successfully")
+            Dialogs.instance.showMsg(R.string.backup_modules_success)
+        } else {
+            Log.e(TAG, "Error copying modules to chosen location.")
+            Dialogs.instance.showErrorMsg(R.string.error_occurred)
+        }
+
+        return ok;
+    }
+
+    suspend fun backupModulesViaIntent(callingActivity: Activity)  = withContext(Dispatchers.Main)   {
+        val fileName = MODULE_BACKUP_NAME
         internalDbBackupDir.mkdirs()
         val zipFile = File(internalDbBackupDir, fileName)
         val books = selectModules(callingActivity) ?: return@withContext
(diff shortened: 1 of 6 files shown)


tuomas2:
After merging other PRs, there are now little merge conflicts

RuggMatt:
Should be good to go now. Also added a commit to remove some of the now unused code...

app/src/main/java/net/bible/android/control/backup/BackupControl.kt
```diff
@@ -113,7 +93,7 @@ class BackupControl @Inject constructor() {
 
         if (ok) {
             Log.d(TAG, "Copied database to chosen backup location successfully")
-            Dialogs.instance.showMsg(R.string.backup_success, uri.path)
```
JJK96, line 116: Why is this removed? I would think it is useful to see where the file was saved.

app/src/main/java/net/bible/android/view/activity/page/MenuCommandHandler.kt
```diff
@@ -220,20 +218,12 @@ constructor(private val callingActivity: MainBibleActivity,
                         .setTitle(callingActivity.getString(R.string.backup_backup_title))
                         .setMessage(callingActivity.getString(R.string.backup_backup_message))
                         .setNegativeButton(callingActivity.getString(R.string.backup_phone_storage)) { dialog, which ->
-                            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.Q) {
```
JJK96, line 223: Since this is removed, I think this needs testing on those android versions.

app/src/main/java/net/bible/android/view/activity/page/MenuCommandHandler.kt
```diff
@@ -259,29 +249,9 @@ constructor(private val callingActivity: MainBibleActivity,
                     requestCode = IntentHelper.UPDATE_SUGGESTED_DOCUMENTS_ON_FINISH
                 }
                 R.id.restore_app_database -> {
-                    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.Q) {
```
JJK96, line 262: Since this is removed, I think this needs testing on those android versions.

JJK96 (changes requested):
I've reviewed the PR. 

It no longer gives an error when trying to save on Android 10, so that's a big improvement already!
I tested saving and restoring the application database and modules to/from phone storage which all worked fine on Android 10.

I ran into one problem, backup to phone storage does not overwrite an existing file. 
So I end up with `andBibleDatabase.db`, `andBibleDatabase.db (1)` and `andBibleDatabase.db (2)`.
That issue has to be resolved before merging, IMO.

app/src/main/java/net/bible/android/control/backup/BackupControl.kt
tuomas2: Check out screenshots on #756.

JJK96: I missed that issue, good change then

app/src/main/java/net/bible/android/view/activity/page/MenuCommandHandler.kt
tuomas2: It would indeed be good to roughly go through different android versions with emulator (from lollipop (5) to the latest models).

RuggMatt:
> 
> 
> I've reviewed the PR.
> 
> It no longer gives an error when trying to save on Android 10, so that's a big improvement already!
> I tested saving and restoring the application database and modules to/from phone storage which all worked fine on Android 10.
> 
> I ran into one problem, backup to phone storage does not overwrite an existing file.
> So I end up with `andBibleDatabase.db`, `andBibleDatabase.db (1)` and `andBibleDatabase.db (2)`.
> That issue has to be resolved before merging, IMO.

Regarding the multiple file names, this is an android file level problem as the intent automatically creates a new file if one with the same name already exists. 

Possible options are to show another screen when choosing backup when user clicks phone storage as described below

 #### Current: 
When a user selects the **phone storage** option:
  - show file picker and write to file
 #### Possible Change:
When a user selects the **phone storage** option
- if it is the user's first time backing up to phone storage
   - ask them where to create a new file
   - write backup to file
   - call [`ContentResolver#takePersistableUriPermission(Uri, int)`](https://developer.android.com/reference/android/content/ContentResolver#takePersistableUriPermission\(android.net.Uri,%20int\)) to keep write permissions to file across phone reboots
   - possibly need to save URI somewhere or figure out how to use [`ContentResolver#getPersistedUriPermissions()`](https://developer.android.com/reference/android/content/ContentResolver#getPersistedUriPermissions\(\))
- else the user has created a backup before
  - ask them if they want to overwrite the old backup file or create a new one?
    - overwrite?
      - retrieve old backup Uri using `getPersistedUriPermissions()` and figure out which one is the correct type of backup?
      - write backup to Uri
      - maybe a menu so a user can select which backup to overwrite if user has created multiple different backups before
    - or create a new file
      - user is prompted to create a new file through the android UI file picker
      - write backup to file
      - `takePersistableUriPermission...`

JJK96:
I like the simplicity of your current solution. I think this is a problem that should be solved by Android indeed.

So, considering that the issue is not easily resolved at the moment, and the current status is an improvement with respect to the previous version, this PR can be merged IMO (After some testing on older android versions).

RuggMatt:
was doing some testing on older versions and i discovered this error in general.. 
When a user backs up the database after opening the app, no "success" message is shown. 
If a user immediately backs up again, the "success" message works as normal. 
```
E/Dialogs: Error showing error message.  Original error msg:My Notes, Bookmarks, Reading Plans and Workspaces backed up successfully.
    java.lang.RuntimeException: Can't create handler inside thread that has not called Looper.prepare()
        at android.os.Handler.<init>(Handler.java:200)
        at android.os.Handler.<init>(Handler.java:114)
        at android.widget.Toast$TN.<init>(Toast.java:344)
        at android.widget.Toast.<init>(Toast.java:100)
        at android.widget.Toast.makeText(Toast.java:258)
        at net.bible.android.view.activity.base.Dialogs.showMsg(Dialogs.kt:119)
        at net.bible.android.view.activity.base.Dialogs.showErrorMsg(Dialogs.kt:87)
        at net.bible.android.view.activity.base.Dialogs.showErrorMsg$default(Dialogs.kt:86)
        at net.bible.android.view.activity.base.Dialogs.showMsg(Dialogs.kt:51)
        at net.bible.android.control.backup.BackupControl.backupDatabaseToUri(BackupControl.kt:96)
        at net.bible.android.view.activity.page.MainBibleActivity$onActivityResult$2.invokeSuspend(MainBibleActivity.kt:1158)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:738)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
```
Does anyone know how to fix?

timbze:
> Does anyone know how to fix?

I great to see this stuff being tested thoroughly. :) This issue is because the dialog is trying to start on a background thread, but it must be started on the main thread.

In `MainBibleActivity.kt`, L1133+ (master branch)
```kotlin
GlobalScope.launch(Dispatchers.IO) {
  val inputStream = contentResolver.openInputStream(data!!.data!!)
  if (backupControl.restoreDatabaseViaIntent(inputStream!!)) {
    windowControl.windowSync.setResyncRequired()
    withContext(Dispatchers.Main) {
      documentViewManager.clearBibleViewFactory()
      currentWorkspaceId = 0
    }
  }
}
```

`restoreDatabaseViaIntent` is being started under Dispatchers.IO, and that is where the error occurs. It either needs to start in Dispatchers.Main, or the Dialogs.instance.showMsg within `restoreDatabaseViaIntent` needs to be within another `GlobalScope.launch(Dispatchers.Main)`

timbze:
@mattschlosser can you either merge or rebase develop into your PR so that CI can pass? There was a CI issue that has been fixed.

JJK96:
I just tested the new version, it works fine for me. 
In the previous version I could no longer choose a filename for the backup to phone storage, but that is fixed in this version.

app/src/main/res/values/strings.xml
```diff
@@ -481,7 +481,8 @@
     <string name="restore_confirmation">Overwrite My Notes, Bookmarks, Reading Plans and Workspaces?</string>
     <string name="restore_success">My Notes, Bookmarks, Reading Plans and Workspaces restored successfully</string>
     <string name="loading_backup">Loading database from backup…</string>
-    <string name="backup_success">My Notes, Bookmarks, Reading Plans and Workspaces backed up successfully to %s.</string>
+    <string name="backup_success">My Notes, Bookmarks, Reading Plans and Workspaces backed up successfully.</string>
```
tuomas2, line 484: Because of translations, we should not make such changes to strings that change parameters. Otherwise users with translations that have not been updated will have crashes.

When parameters need to be changed, like here, we need to introduce a new string (with new key, for example backup_success2), use it, and remove old string.

tuomas2:
Looks okay. I will do the fix related to my own comment in develop branch. Let's continue testing there.

PR #787, opened 2020-07-14
Handle alerts properly when user clicks outside alert area or pushes back button

RuggMatt:
This handles some edge cases on alert dialogs. Some dialogs did not handle the case when a user hits the back button or clicks the area around/behind the dialog. If the dialog was in a co-routine, this lead to the co-routine never resuming, and the activity may appear "stuck"

For example, if a user happened to click the area around/behind an alert, or pushes the phone's back button, when seeing the download warning, the user would see a download loading spinner forever, as they never explicitly allowed or denied downloads. This fixes this case by associating an outside click/back button action as the same as hitting the negative button.

I searched through the whole code base and these are the 5 I thought where we may need to handle these cases, as they were all in suspend functions/co-routines

**Screenshots**

### DIFF
--- a/app/src/main/java/net/bible/android/control/backup/BackupControl.kt
+++ b/app/src/main/java/net/bible/android/control/backup/BackupControl.kt
@@ -214,6 +214,7 @@ class BackupControl @Inject constructor() {
                     }
                     .setNeutralButton(R.string.select_all) { _, _ -> it.resume(null) }
                     .setNegativeButton(R.string.cancel) { _, _ -> it.resume(null) }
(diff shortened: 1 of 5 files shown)


timbze (commented):
One other option is to use `setCanceledOnTouchOutside(false)`

But not sure if it's needed here. Generally pressing outside of dialog means Cancel for the user, I think. Unless it's by mistake.

Edit:
Even if using my option here, I think will still need to use `.setOnCancelListener` that Matt implemented here to handle back press

tuomas2:
Good catch @mattschlosser! @timbze's comment need to be considered case-by-case. I can check this when merging the code.

PR #793, opened 2020-07-16
Update README.md

Poussinou:
Hi,

This small PR adds two buttons to get your app on F-Droid and Google Play 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,13 @@
 
 Powerful offline Bible Study tool for Android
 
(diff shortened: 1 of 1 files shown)


tuomas2:
@Poussinou if you would be interested to contribute to othet marketing material please send me email:)

Poussinou:
@tuomas2 Thanks for the proposition, but sorry nope, I'm just putting the badge(s) when an application come to F-Droid (and adds the Google Play one when not already there) ;)

PR #796, opened 2020-07-18
Fixes #736 - window button misplaced in full screen

### LINKED ISSUE #736: Window button misplaced in full screen
DSC_5967
*Sent by Tuomas Airaksinen. Created by [fire](https://fire.fundersclub.com/).*

---
huawei p30 pro

---
*Attachments:*

1. [DSC_5967.JPG](https://firebot-prod-media.s3.amazonaws.com:443/email-attachments/a1dff61f-cd55-48af-92f3-2e7de1e2f292/DSC_5967.JPG)

RuggMatt:
Closes #736. 

Also fixed animations when entering/exiting full screen. 

It seems when entering full screen, the button was being destroyed and then recreated. This led to no animation. 

On exiting full screen, padding was not being set correctly, due to a lingering reference to a finished ActionMode. Once that was fixed, the animation after exiting fullscreen was delayed because `toggleWindowButtonVisibility` was not being called after a `FullScreenEvent`,  and would wait the 2 seconds until the touch delay to fade the buttons was fired. 

Demo
output

**Screenshots**


tuomas2:
Bugs introduced:
 - [x] when changing "hide window buttons" from application preferences, change is not immediately applied, but only after new window is created etc. Actually only those windows that are refreshed will have button (for example, creating a window will create a window that has a window button).

Not sure if new bug:
- [x] when in action mode, you can double-tap to toggle fullscreen on/off
- [x]  when entering action mode in full screen, window separator disappears for a moment (it comes back after little delay)

RuggMatt:
still working on this...

i fixed the bugs you described, but i have found new ones since then:
- [x] the window button is showing when a window is maximized

tuomas2:
> * the window button is missing when opening the application and a window is maximized.

Hmm, I'm not quite sure if we should consider this a bug as most of items from window button menu are not accessible anyway in maximize mode. You can access the remaining items (synchronize and text options) by long-clicking unmaximize button, but most users don't need this.

RuggMatt:
@tuomas2 Yes. The issue was the button was not hiding correctly when the window was first maximized. 

I think I have finished on this fix. If you find any more bugs, let met know.

tuomas2:
Seem to work now. Will merge and test for some days before deploying.

PR #839, opened 2020-09-07
Closes #832 - Window buttons jump when coming back from find

### LINKED ISSUE #832: Window button bar jumps when coming back from find
1. Go to "find" view from main menu
2. Type something
3. When keyboard is still active, press 'up' button from toolbar
 -> Goes back to main bible view (OK)
 -> window button bars jump weirdly to up and down, height of keyboard or so (NOK)

Reproduced in develop branch (250be8045f6409625af53d7a26c71e1221a098cb, build beta-390.1).
Can't reproduce this on emulator. Reproduces on my Samsung tablet quite consistently, but not always.

RuggMatt:
Closes #832 

Added a method to close the keyboard before exiting the find activity so a
resize event is not triggered in the main activity, which prevents the window
buttons from jumping.

### DIFF
--- a/app/src/main/java/net/bible/android/view/activity/base/ActivityBase.kt
+++ b/app/src/main/java/net/bible/android/view/activity/base/ActivityBase.kt
@@ -227,7 +227,7 @@ abstract class ActivityBase : AppCompatActivity(), AndBibleActivity {
             val inputMethodManager: InputMethodManager = getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
             inputMethodManager.hideSoftInputFromWindow(currentFocus!!.windowToken, 0)
         } catch (e: Exception) {
-            Log.e("AndroidView", "closeKeyboard: $e")
+            Log.e(TAG, "closeKeyboard: $e")
         }
     }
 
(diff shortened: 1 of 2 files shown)


timbze:
I think a "universal" solution for this is probably ok. Is there an error being thrown if keyboard is not open or in some cases? I think the try .. catch is fine for that part, unless an error is constantly being thrown then it would be better to figure out another way. If it's really just a safety then that's fine.

RuggMatt:
I found this solution to close the keyboard and it had the try/catch there. I don't know if it is necessary. It does not throw if the keyboard isn't open. So more for safety.

timbze:
> I found this solution to close the keyboard and it had the try/catch there. I don't know if it is necessary. It does not throw if the keyboard isn't open. So more for safety.

I think it's good then

app/src/main/java/net/bible/android/view/activity/base/ActivityBase.kt
```diff
@@ -217,6 +218,17 @@ abstract class ActivityBase : AppCompatActivity(), AndBibleActivity {
...
+        closeKeyboard()
+    }
+
+
+    private fun closeKeyboard() {
+        try {
+            val inputMethodManager: InputMethodManager = getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
+            inputMethodManager.hideSoftInputFromWindow(currentFocus!!.windowToken, 0)
+        } catch (e: Exception) {
+            Log.e("AndroidView", "closeKeyboard: $e")
```
tuomas2:
```suggestion
            Log.e(TAG, "closeKeyboard: $e")
```

PR #844, opened 2020-09-09
show MyNotes as a window

JJK96:
MyNotes are shown as a window in bible view.
The notes are wrapped in 
```
<div id="editor">
</div>
```
When the edit button is pressed, the div is made `contentEditable` and the content is selected.

When the user clicks the save button the `saveNote` method on `BibleJavascriptInterface.kt` is called in order to save the note.

In the future, similar techniques can be used to make the bible view editable in order to have more flexible markup.

# Possible problems

1. In order to add the bibleview javascript to the notes page, I use the `HtmlMessageFormatter` which is actually intended for error messages, but it works. Possibly we could rename the HtmlMessageFormatter to unlink it from error messages, since it works fine with other types of content.
~2. The notes are saved on every keypress, on my hardware this is not noticeable, but on some devices this inefficiency may be noticed. I currently do not see a solution to save only when done editing since there's only the `input` event to detect a change, and not an event when the input window is closed.~

# TODO

- [x] Clean up files belonging to the old MyNote editing method.
- [x] Synchronization with other window, automatically get the correct note for the verse that is currently viewed.
- [x] Edit and save only when the respective buttons are pressed
- [x] Reload when note has changed (by editing in other window)
- [ ] Rich text editing

Please have a look at these problems and give me some feedback on the right approach.


app/src/main/java/net/bible/android/control/page/CurrentMyNotePage.kt
```diff
@@ -123,6 +133,11 @@ class CurrentMyNotePage internal constructor(
         doSetKey(key)
 	}
 
+    @JavascriptInterface
+    override fun save(text: String) {
+        myNoteControl.saveMyNoteText(text)
+    }
+
```
tuomas2: Perhaps better that we have all js interface functions in net.bible.android.view.activity.page.BibleJavascriptInterface. Or what do you think?

app/src/main/java/net/bible/android/view/activity/page/BibleView.kt
```diff
@@ -279,6 +281,8 @@ class BibleView(val mainBibleActivity: MainBibleActivity,
             finalHtml = finalHtml.replace("<div id='start'>", "<div id='start' style='height:${startPaddingHeight}px'>")
 
             val currentPage = window.pageManager.currentPage
+            this.settings.javaScriptEnabled = true
```
tuomas2: already on line 191.

app/src/main/java/net/bible/android/view/activity/page/BibleView.kt
```diff
@@ -279,6 +281,8 @@ class BibleView(val mainBibleActivity: MainBibleActivity,
...
+            this.addJavascriptInterface(currentPage, "currentPage")
```
tuomas2: see line 148 (setBibleJavascriptInterface)

app/src/main/java/net/bible/android/view/activity/page/BibleView.kt
```diff
@@ -30,6 +30,7 @@ import android.webkit.JsResult
 import android.webkit.WebChromeClient
 import android.webkit.WebView
 import android.webkit.WebViewClient
+import androidx.appcompat.widget.AppCompatEditText
```
tuomas2: this isn't used here?

app/src/main/java/net/bible/android/view/activity/page/BibleView.kt
```diff
@@ -56,6 +57,7 @@ import net.bible.android.view.activity.page.screen.WebViewsBuiltEvent
 import net.bible.android.view.util.UiUtils
 import net.bible.service.common.CommonUtils
 import net.bible.service.device.ScreenSettings
+import net.bible.service.format.HtmlMessageFormatter
```
tuomas2: this isn't used here?

app/src/main/java/net/bible/android/control/page/CurrentMyNotePage.kt
JJK96: I only discovered later that there was a JavaScript interface already, I'll move it over to there.

JJK96:
This PR is now ready for review/merge, I have been using this for a while and it performs the editing of notes correctly.

The only improvement I see currently is that edit mode could be enabled when creating a new note, but I don't see how to do that right now and I don't think it's a hard requirement.

Rich editing is still open in the to do, but that's more of a new feature, since the current mynote editing also doesn't have rich editing.

app/src/main/java/net/bible/android/control/mynote/MyNoteControl.kt
```diff
@@ -134,6 +135,11 @@ class MyNoteControl @Inject constructor(val activeWindowPageManagerProvider: Act
             text = myNote.noteText
             if (abbreviated) { //TODO allow longer lines if portrait or tablet
                 val singleLine = true
+                if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.N) {
+                    text = Html.fromHtml(text, Html.FROM_HTML_MODE_LEGACY).toString();
+                } else {
+                    text = Html.fromHtml(text).toString();
+                }
```
tuomas2:
```suggestion
                text = if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.N) {
                    Html.fromHtml(text, Html.FROM_HTML_MODE_LEGACY).toString();
                } else {
                    Html.fromHtml(text).toString();
                }
```

app/src/main/java/net/bible/android/control/page/CurrentMyNotePage.kt
```diff
@@ -18,10 +18,13 @@
 package net.bible.android.control.page
 
 import android.util.Log
+import android.webkit.JavascriptInterface
```
tuomas2:
```suggestion
```

app/src/main/java/net/bible/android/control/page/CurrentMyNotePage.kt
```diff
@@ -38,19 +41,25 @@ import java.io.IOException
...
     // just one fake book for every note
     private var fakeMyNoteBook: Book? = null
     private var fakeMyNoteBookVersification: Versification? = null
-    override val currentPageContent: String get() = myNoteDAO.getMyNoteTextByKey(key)
+    override val currentPageContent: String get() {
+        var text: String = myNoteDAO.getMyNoteTextByKey(key)
+        text = text.replace("\n", "<br>")
+        text = "<div id=\"editor\">$text</div>"
+        text = HtmlMessageFormatter.format(text)
+        return text
```
tuomas2:
```suggestion
        val text= myNoteDAO.getMyNoteTextByKey(key).replace("\n", "<br>")
        return HtmlMessageFormatter.format("<div id=\"editor\">$text</div>")
```

app/src/main/java/net/bible/android/view/activity/page/BibleView.kt
```diff
@@ -125,6 +125,8 @@ class BibleView(val mainBibleActivity: MainBibleActivity,
             windowRef = WeakReference(value)
         }
 
+    var isEditing = false
```
tuomas2, line 130:
```suggestion
    // editing mode for MyNotes
    var isEditing = false
```

app/src/main/java/net/bible/android/view/activity/page/MainBibleActivity.kt
```diff
@@ -159,6 +159,11 @@ class MainBibleActivity : CustomTitlebarActivityBase(), VerseActionModeMediator.
             if (::documentControl.isInitialized) {
                 documentControl.isMyNotes
             } else false
+    val isEditing: Boolean
+        get() {
+            var bibleview = windowControl.activeWindow.bibleView
+            return bibleview != null && bibleview.isEditing
```
tuomas2:
```suggestion
            val bibleView = windowControl.activeWindow.bibleView
            return bibleView != null && bibleView.isEditing
```

app/src/main/java/net/bible/android/view/activity/page/MainBibleActivity.kt
```diff
@@ -159,6 +159,11 @@ class MainBibleActivity : CustomTitlebarActivityBase(), VerseActionModeMediator.
             if (::documentControl.isInitialized) {
                 documentControl.isMyNotes
             } else false
+    val isEditing: Boolean
```
tuomas2:
```suggestion

    private val isEditing: Boolean
```

app/src/main/java/net/bible/android/view/activity/page/screen/DocumentViewManager.kt
```diff
@@ -82,34 +79,25 @@ class DocumentViewManager @Inject constructor(
...
 
-    private fun getDocumentView(window: Window): DocumentView? {
-        return if (myNoteViewBuilder.isMyNoteViewType) {
-            myNoteViewBuilder.view
-        } else { // a specific screen is specified to prevent content going to wrong screen if active screen is changed fast
-            splitBibleArea?.bibleViewFactory?.getOrCreateBibleView(window)
-        }
+    fun getDocumentView(window: Window): DocumentView? {
+        // a specific screen is specified to prevent content going to wrong screen if active screen is changed fast
+        return splitBibleArea?.bibleViewFactory?.getOrCreateBibleView(window)
```
tuomas2:
```suggestion
    private fun getDocumentView(window: Window): DocumentView? =
        splitBibleArea?.bibleViewFactory?.getOrCreateBibleView(window)
```

app/src/main/res/values/strings.xml
```diff
@@ -472,6 +472,7 @@
     <string name="mynotes">My Notes</string>
     <string name="mynote_add">Add My Note</string>
     <string name="mynote_saved">My Note saved</string>
+    <string name="save">Save</string>
```
tuomas2: Better make strings contextual, i.e. id should be mynote_save

tuomas2:
- [ ] I'm wondering if edit mode is needed. It's confusing to me, although I think I finally figured out how it is supposed to work. It could be always in edit mode and with on-blur event (or periodically, after few seconds of editing stops) save it.
- [ ] entering "my notes" mode is not very user friendly. I'm wondering if that should appear as a "commentary" in commentary list (both in commentary button, as well as document selector), then it might be intuitive, as it now functions more or less similar way as commentaries anyway. But now we need to go there through action mode.
- [ ] when in edit mode, keyboard easily overlaps the window. Could try to use panning mode for virtual keyboard?
- [x] When there's no text, it should be indicated somehow (for example text "Start editing" with gray in text box, which disappears when starting editing). Blank window looks always like a bug.
- [x] Clicking/long-clicking pen button (within bible text) should probably function in this new design similar way as clicking links (to "special window" / context menu/ other options, according to settings)

app/src/main/java/net/bible/android/view/activity/page/screen/DocumentViewManager.kt
JJK96: In the future this function could become more complex again, so I'm not in favour of this change.

app/src/main/res/values/strings.xml
JJK96: When I was translating I noticed that I was processing a lot of similar messages. 
So, I thought it was better to have only one "save" message, instead of multiple.

JJK96:
- In the previous version I did not have the edit mode. I added it because since notes are in a window now, we could have a workflow where you just view notes in a separate window while you are reading a text. Suppose in the future we have bible references in the notes (which I intend to add soon), we want to have those clickable, which doesn't work in edit mode.
I agree that the edit mode is not very intuitive, especially when creating a new note, since you need to do an extra action, we could think of a way to improve that by going into edit mode by default when a new note is created.

> entering "my notes" mode is not very user friendly.
- This might be a valid point, but it's not related to this PR I think, since it is that way in the `develop` branch as well.

app/src/main/java/net/bible/android/view/activity/page/screen/DocumentViewManager.kt
tuomas2: If a function is one-liner that returns a value, just press alt+enter on top of "return" and it offers to change it to "expression body" form. And as easily  it can be changed back to "body block" form if one day it needs to be made more complex. Another thing here was that I believe the comment is outdated or otherwise unclear, so better remove it.

app/src/main/res/values/strings.xml
tuomas2: True, but it is possible that in other languages "save" is translated differently in different contexts. That's why I think it's safer to have repetition in translation strings.

tuomas2:
> * In the previous version I did not have the edit mode. I added it because since notes are in a window now, we could have a workflow where you just view notes in a separate window while you are reading a text. Suppose in the future we have bible references in the notes (which I intend to add soon), we want to have those clickable, which doesn't work in edit mode.
>   I agree that the edit mode is not very intuitive, especially when creating a new note, since you need to do an extra action, we could think of a way to improve that by going into edit mode by default when a new note is created.

I see. Maybe changing to edit mode could be with a button that is within the window? Or from window button menu?

> > entering "my notes" mode is not very user friendly.
> 
> * This might be a valid point, but it's not related to this PR I think, since it is that way in the `develop` branch as well.

It's true, but before releasing a new version we should think something that works well. So that's why I think it's better to continue working on this feature until we are fully satisfied with the feature and UX and believe that users are also as little confused as possible. So I would highly suggest that we do this change here too. I can help if needed. 

If you want to split development to smaller pieces, we could consider merging this, but with that in mind, that we won't release until other parts are also ready.

JJK96:
> when in edit mode, keyboard easily overlaps the window. Could try to use panning mode for virtual keyboard?

I cannot get this to work. Also, I do not consider this a problem, since you can easily scroll up.

JJK96:
Now when the note is empty we show  a placeholder.
When the placeholder is clicked we enter editing mode.

On blur, we leave editing mode and save the note.

JJK96:
This can be closed, all functionality is in the develop branch now

Issue #847, opened 2020-09-15
Window buttons not displayed when opened in split screen


tuomas2:
1. Open some other app in upper screen
2. Open And Bible in lower screen
 -> window button bar will go under system nav bar

This is reproduced with my samsung tablet @ android 10. 3d0df998, dev branch (build 390.1 or so).
Probably does not reproduce with 3.3.

tuomas2:
@mattschlosser this might be related to the changes you have made recently

RuggMatt:
I can also reproduce. :+1:

RuggMatt:
Also can reproduce on 3.3.390 (Play store)

tuomas2:
Seem to be only on Android 10 (and probably on more recent as well).

PR #848, opened 2020-09-17
Ensure window button is visible when in multi window mode

RuggMatt:
Work for #847 

- Window buttons can be seen when in spilt screen mode

- Not sure why this check was here.. @tuomas2 ?
-  It seems that the window buttons jump when entering/exit full screen now...

**Screenshots**


tuomas2:
>Not sure why this check was here.. @tuomas2 ? 

Hmm strange. I checked with blame and found that it is fix to #725. Is this now something that behaves differently on different android versions or something? Because since #725, my tablet has updated from Android 9 to 10.

tuomas2:
Yes, it looks like we need to make if for >= API 29.

tuomas2:
Thanks for helping investigating this. Committed fix 6c4d38018e29e97f6a19fc276554866933d6acf0.

PR #849, opened 2020-09-21
ACRA bug reporting

timbze:
Here is the basic of what is required for ACRA bug reporting. This is for information and can be closed if not interested in it.

Optionally, it is possible to also specify if crash report should be sent automatically. Or to give the user the option to send it and add a note if the user wants to.

This feature adds about .06 mb to release apk file

This feature also requires our own backend. See more https://github.com/ACRA/acra/wiki/Backends

Link to ACRA: https://github.com/ACRA/acra

### DIFF
--- a/app/bibleview-js/package-lock.json
+++ b/app/bibleview-js/package-lock.json
@@ -3886,9 +3886,9 @@
       }
     },
     "npm": {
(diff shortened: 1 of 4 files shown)


tuomas2:
#852 replaces this

PR #852, opened 2020-09-24
Use ACRA email reporter to send even better bug reports

tuomas2:
Does not require new strings, thus this is towards current-stable, i.e. planning to release it in 3.3 bugfix release.
Uses "Report a bug" as email title, can be fixed in 3.4 to something better such as "Bug report for And Bible". Also email body has one cosmetic problem which requires changing string, i.e. there's "%s" in the end of the message. 

Example report: 

https://support.andbible.org/scp/tickets.php?id=101


tuomas2:
Closes #849

PR #854, opened 2020-09-25
Send report as file and show more lines from logcat

### LINKED ISSUE #853: Acra error emails are empty
tested with build 392 (develop). not releasing 392 to play store before this is fixed.

timbze:
This is one way to fix #853


tuomas2:
Hmm seem to work on my mobile phone just by using attachment. Maybe Gmail does not like too big body text. But it's not very practical that everything is in attachment.

timbze:
I thought it was perhaps better for end user if text in email was shorter... But it is an extra step to download the file and then open when making use of the information.

tuomas2:
Reverted Acra for now and created #855

PR #870, opened 2020-10-13
Feature/room for bookmark

tuomas2:
I would still like to
 - [x] get rid of BookmarkDto and LabelDto (replace with entities)
 - [ ] make it possible to add multiple bookmarks in one verse (for different labels)


app/src/main/java/net/bible/service/db/DatabaseContainer.kt
```diff
@@ -551,6 +551,30 @@ private val SQUASH_30_33 = object : Migration(30, 33) {
...
+            execSQL("CREATE INDEX IF NOT EXISTS `index_BookmarkToLabel_labelId` ON `BookmarkToLabel` (`labelId`)")
+
+            execSQL("INSERT INTO Bookmark SELECT * from bookmark_old;")
+            execSQL("INSERT INTO Label SELECT * from label_old;")
+            execSQL("INSERT INTO BookmarkToLabel SELECT * from bookmark_label;")
+
+            execSQL("DROP TABLE bookmark_old;")
+            execSQL("DROP TABLE label_old;")
+            execSQL("DROP TABLE bookmark_label;")
+            execSQL("PRAGMA foreign_keys=ON;")
```
timbze: Possibly we should stop using the `PRAGMA foreign_keys=ON or OFF` to prevent further issues of https://github.com/AndBible/and-bible/issues/770#issuecomment-658889797

I don't think it will be an issue here, but since Sqlite 3.26+ operates different on foreign keys with table rename, if we would stop using `PRAGMA foreign_keys` then we would find the issue with earlier Sqlite versions too (i.e. then earlier versions of Sqlite would operate the same way as 3.26+).

db/src/main/java/net/bible/android/database/BookmarksDao.kt
```diff
@@ -0,0 +1,127 @@
...
+    fun allLabels(): List<BookmarkEntities.Label>
+
+    @Query("SELECT * from bookmark")
+    fun allBookmarksSorted(): List<BookmarkEntities.Bookmark>
+
+    @Query("SELECT * from Bookmark where id = :bookmarkId")
+    fun bookmarkById(bookmarkId: Long): BookmarkEntities.Bookmark
+
+    @Query("SELECT * from Bookmark where id IN (:bookmarkIds)")
+    fun bookmarksByIds(bookmarkIds: LongArray): List<BookmarkEntities.Bookmark>
```
timbze: You can also do `List<Long>` I believe. I've done `List<Int>`

tuomas2: Yes; I just used this as it is simpler because the data is coming in this form where this function is used

app/src/main/java/net/bible/service/db/DatabaseContainer.kt
tuomas2: Okay here it indeed works without disabling foreign keys. So removed it.

timbze: My point is that after sqlite 3.26+ it is useless to disable foreign keys and all it does is possibly cause it to work for < 3.26 but give issues for 3.26+. So in my opinion we should make it work without disabling foreign keys from here on, so that there will not be the issue at 3.26+

tuomas2: So we should actually remove use of it from all migrations, rather than just this one?
But sometimes it is needed to remove foreign keys handling when migrating data, isn't it? 
Especially if the data has links to the same table or if two tables have both links to each other.

timbze: I would not remove from previous migrations.

You're right, I did not think about all circumstances. For example, when changing data on only 1 table that has links from other tables it could be needed.

It's just the way foreign keys are handled differently after 3.26 that we need to keep in mind for migrations, but probably won't usually be an issue.

app/src/main/java/net/bible/android/control/bookmark/BookmarkControl.kt
```diff
@@ -56,293 +60,191 @@ open class BookmarkControl @Inject constructor(
...
+        var v = verse
         if (v.verse == 0) {
             v = Verse(v.versification, v.book, v.chapter, 1)
         }
-        val bookmarkDto = getBookmarkByKey(v)
-        if (bookmarkDto?.playbackSettings != null) {
-            bookmarkDto.playbackSettings = settings
-            addOrUpdateBookmark(bookmarkDto)
-            Log.d("SpeakBookmark", "Updated bookmark settings " + bookmarkDto + settings.speed)
+        val bookmark = firstBookmarkStartingAtVerse(v)
```
tuomas2: this should be speak bookmark

app/proguard-rules.pro
```diff
@@ -91,3 +91,25 @@
...
+# kotlinx-serialization-json specific. Add this if you have java.lang.NoClassDefFoundError kotlinx.serialization.json.JsonObjectSerializer
+-keepclassmembers class kotlinx.serialization.json.** {
+    *** Companion;
+}
+-keepclasseswithmembers class kotlinx.serialization.json.** {
+    kotlinx.serialization.KSerializer serializer(...);
+}
+
+# Change here com.yourcompany.yourpackage
+-keep,includedescriptorclasses class com.yourcompany.yourpackage.**$$serializer { *; } # <-- change package name to your app's
```
timbze: Looks like com.yourcompany... needs to be changed at a few lines

JJK96:
It seems that old bookmarks are deleted from the database, ~I haven't looked into it close enough to know if this is intentional or not.~
Anyway, they should either be preserved or converted to a new format if necessary.

- [x] EDIT: I see that the intention is to convert the bookmarks, but it is not working on my phone/database.

JJK96:
screencap

- [x] This menu should display the labels with the style applied. I assume how it is now is just for testing.

tuomas2:
You can send logcat so we can investigate the migration issue (or just do normal bug report via main menu -> report a bug)

tuomas2:
or alternatively send your app db backup

tuomas2:
> This menu should display the labels with the style applied. I assume how it is now is just for testing.

no, it's a bug :-)

timbze:
Updated my AB to latest version with this upgrade and everything seems to work great

PR #880, opened 2020-10-23
Allow any workspace to be deleted at any time, except if it is the only one.

### LINKED ISSUE #836: Should be able to delete current workspace
Now must switch first to another workspace.

RuggMatt:
Closes #836. 

It seems to respond correctly if deleting the active workspace already by switching to another workspace on save. Just changed the delete menu item to be disabled if it's the only workspace, as I'm not sure what would happen then?

**Screenshots**

### DIFF
--- a/app/src/main/java/net/bible/android/view/activity/workspaces/WorkspaceSelectorActivity.kt
+++ b/app/src/main/java/net/bible/android/view/activity/workspaces/WorkspaceSelectorActivity.kt
@@ -465,7 +465,7 @@ class WorkspaceSelectorActivity: ActivityBase() {
         val menu = popup.menu
         popup.menuInflater.inflate(R.menu.workspace_popup_menu, menu)
         val delItem = menu.findItem(R.id.deleteWorkspace)
-        delItem.isEnabled = workspaceEntity.id != windowControl.windowRepository.id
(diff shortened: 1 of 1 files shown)


app/src/main/java/net/bible/android/view/activity/workspaces/WorkspaceSelectorActivity.kt
```diff
@@ -465,7 +465,7 @@ class WorkspaceSelectorActivity: ActivityBase() {
         val menu = popup.menu
         popup.menuInflater.inflate(R.menu.workspace_popup_menu, menu)
         val delItem = menu.findItem(R.id.deleteWorkspace)
-        delItem.isEnabled = workspaceEntity.id != windowControl.windowRepository.id
+        delItem.isEnabled = dataSet.size != 1 // cannot delete the only workspace
```
timbze: This should always work, but it might be better practice to use "> 1" instead?

RuggMatt:
```suggestion
        delItem.isEnabled = dataSet.size > 1 // cannot delete the only workspace
```

timbze (approved):

PR #888, opened 2020-11-12
Feature/vuecli -- new vue3 bibleview-js frontend

tuomas2:
Known issues (that will be fixed after merge, but before next stable release):

- [ ] Performance in general due to vue.js rendering. Solution: render smaller chunks at time and rest asynchronously without disturbing user. 
- [x] Performance of bookmarking long verse ranges. Solution: use different routine for those parts that contain whole verses.
- [ ] Speak bookmarks are not displayed correctly
- [ ] ensureBibleViewVisible() - when writing notes, text input could be hidden
- [ ] readFootnotesAndReferences() - list view of these is not implemented (thinking if it is important or not)
- [ ] compare translations (list view) function missing
- [ ] share / copy function missing
- [ ] Custom CSS not implemented (@bvahedy note!!!)
- [ ] Debug / production mode for vuecli builds automatically in build.gradle
- [ ] LTR texts
- [ ] translation of strings (some js i18n lib needed)
- [ ] "Show bookmarks" should be removed or moved under "Bookmarking settings"
- [ ] Jump to offset (for non-bibles) does not work

Note:
- bookmark styles like underline and "yellow star" are not supported any more


db/src/main/java/net/bible/android/database/WorkspaceEntities.kt
```diff
@@ -40,7 +41,7 @@ val json = Json {
 class WorkspaceEntities {
     data class Page(
         val document: String?,
-        val key: String?,
+        val key: Key?,
```
timbze, line 44: Looks like you might need to migration for this yet? I get crash when trying to open app:
```
2020-11-28 11:02:02.643 6858-6858/net.bible.android.activity I/MainBibleActivity: Creating MainBibleActivity
2020-11-28 11:02:02.649 6858-6858/net.bible.android.activity I/net.bible.android.view.activity.page.MainBibleActivity: onCreate:net.bible.android.view.activity.page.MainBibleActivity@131fae6
2020-11-28 11:02:02.692 6858-6858/net.bible.android.activity D/SpeakSettings: SpeakSettings loaded! SpeakSettings(synchronize=true, replaceDivineName=false, autoBookmark=false, restoreSettingsFromBookmarks=false, playbackSettings=PlaybackSettings(speakChapterChanges=true, speakTitles=true, speakFootnotes=false, speed=100, bookId=null, bookmarkWasCreated=null, verseRange=null), sleepTimer=0, lastSleepTimer=10, queue=true, repeat=false, numPagesToSpeakId=0)
2020-11-28 11:02:02.776 6858-6858/net.bible.android.activity D/net.bible.android.control.navigation.DocumentBibleBooksFactory: Initialising DocumentBibleBooksFactory cache
2020-11-28 11:02:02.808 6858-6858/net.bible.android.activity D/WinRepository: onLoadDb for workspaceId=1
2020-11-28 11:02:02.809 6858-6858/net.bible.android.activity D/WinRepository: clear false
2020-11-28 11:02:02.811 6858-6858/net.bible.android.activity D/AndroidRuntime: Shutting down VM
2020-11-28 11:02:02.812 6858-6858/net.bible.android.activity E/AndroidRuntime: FATAL EXCEPTION: main
    Process: net.bible.android.activity, PID: 6858
    java.lang.RuntimeException: Unable to start activity ComponentInfo{net.bible.android.activity/net.bible.android.view.activity.page.MainBibleActivity}: java.lang.IllegalArgumentException: bad base-64
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3449)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3601)
        at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:85)
        at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
        at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2066)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:223)
        at android.app.ActivityThread.main(ActivityThread.java:7656)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)
     Caused by: java.lang.IllegalArgumentException: bad base-64
        at android.util.Base64.decode(Base64.java:163)
        at android.util.Base64.decode(Base64.java:138)
        at android.util.Base64.decode(Base64.java:120)
        at net.bible.android.database.Converters.strToKey(AppDatabase.kt:102)
        at net.bible.android.database.WorkspaceDao_Impl.pageManager(WorkspaceDao_Impl.java:2508)
        at net.bible.android.control.page.window.WindowRepository.loadFromDb(WindowRepository.kt:377)
        at net.bible.android.control.page.window.WindowRepository.initialize(WindowRepository.kt:110)
        at net.bible.android.view.activity.page.MainBibleActivity.onCreate(MainBibleActivity.kt:222)
        at android.app.Activity.performCreate(Activity.java:8000)
        at android.app.Activity.performCreate(Activity.java:7984)
```

tuomas2: yes... I'm still not sure if I leave it like this. base64 binary serialization is not so desirable option.
Though this affects only dictionaries and gen books.

tuomas2: I was considering also to use Gson. But it's bit more difficult.

timbze:
In night mode the text is invisible yet. I guess not changing color

tuomas2:
> In night mode the text is invisible yet. I guess the text is not changing color

Yeo, I guess it's not implemented yet.

timbze:
- [ ] In night mode the text is still not correct when night text color is set to default. (red letters are ok)
- [ ] I think there should be a way to easily bookmark a verse without having to highlight it all manually. 
- [ ] IMO, Verse number is not needed at start of commentaries (see pic 1 below)
- [ ] I think the bookmark "create**d** in" should have short Bible name, and not Osis ID. This should show bookmark was made in KJV  (see pic 2 below)

**Picture 1:**
image

**Picture 2:**
Screenshot from 2021-01-18 12-16-40

tuomas2:
> * [ ]  I get this error on migration. When I delete database so that migrations don't run then it works.

Is this fatal or does it have consequences? Something is not working?

timbze:
> > * [ ]   I get this error on migration. When I delete database so that migrations don't run then it works.
> 
> Is this fatal or does it have consequences? Something is not working?

Yes, migration fails and app fails to start because Room says that database does not match schema

tuomas2:
> Yes, migration fails and app fails to start because Room says that database does not match schema

Can you send your DB to me then for debugging.

tuomas2:
> Yes, migration fails and app fails to start because Room says that database does not match schema

It's inside try-catch block and it should just log error and not fail, so it's weird that schema is left wrong. I would expect missing notes as a consequence.

timbze:
> > Yes, migration fails and app fails to start because Room says that database does not match schema
> 
> It's inside try-catch block and it should just log error and not fail, so it's weird that schema is left wrong. I would expect missing notes as a consequence.

You're right. There must be some other issue yet, because the error is that Bookmark table is not migrated correctly. Let me see if I can pinpoint where that comes from

tuomas2:
Moved leftovers to #922

PR #915, opened 2021-01-03
Add Flipper dev tool to debug configuration

timbze:
Flipper dev tool is upgrade from stetho. Has more and better features

Had to upgrade gradle for CI to pass


timbze:
~~Don't know what the issue is with CI, need to check it out a bit more first~~

PR #930, opened 2021-02-07
Prevent crashing on restoring a window with unknown document

JJK96:
If documentCategory was "Other" in the database, restoring would crash the application.
Now it will just set another document type for the window.


tuomas2:
Pushed a little modified fix

PR #932, opened 2021-02-10
Draft: Fix crash when displaying a bookmark with endOffset too high

JJK96:
This is a quick fix, we need to debug why a bookmark was created with an endOffset of 1 higher than the verse lenght.

Also, I should probably make it > instead of >=


tuomas2:
I believe that I fixed this in journal branch where I found this originally

PR #937, opened 2021-02-28
Refactor/kotlin android extensions deprecated

timbze:
Kotlin-android-extensions plugin is deprecated

This PR is to migrate Kotlin synthetic view binding to Jetpack layout view binding so that the deprecated plugin can be removed

https://developer.android.com/topic/libraries/view-binding/migration

### Issues
There are no issues in current PR with below items. These are more complex to deal with in migrating away from kotlin-android-extensions.
~- [ ] DocumentListItem.kt~
~- [ ] MyNoteItemAdapter.kt (uses Android's TwoLineListItem which does not have a binding, will probably need to create a layout for it)~
~- [ ] BookmarkItemAdapter.kt~
~- [ ] SpeakTests.kt~

### Need to test
- [x] First startup download document activity, that OK button shows
- [x] History activity

### Other
- Added back (up) button to DailyReadingList and ReadingPlanSelectorList


timbze:
This PR could be merged after satisfactory review.

There is more work to be done in migrating away from kotlin-android-extensions but that doesn't prevent this from being merged.

Edit: Don't know what the preference is, but I could squash the commits if that is preferred.

tuomas2:
> This PR could be merged after satisfactory review.
> 
> There is more work to be done in migrating away from kotlin-android-extensions but that doesn't prevent this from being merged.
> 
> Edit: Don't know what the preference is, but I could squash the commits if that is preferred.

I prefer separate commits, keep it like this

tuomas2 (approved):
Looks good! Thank you!

PR #969, opened 2021-04-12
EditorConfig set imports to not use wildcard (e.g. java.io.*)

timbze:


.editorconfig
```diff
@@ -9,9 +9,13 @@ charset=utf-8
...
 
 [*.kt]
 indent_style=space
 indent_size=4
+
+[*.{kt, java}]
 max_line_length=120
+ij_kotlin_name_count_to_use_star_import = 999
+ij_kotlin_name_count_to_use_star_import_for_members = 999
+ij_java_class_count_to_use_import_on_demand = 999
```
tuomas2, line 21: there is not "disabled" option, like 0 or -1?

timbze: Android Studio has the setting but I could not find that there is such a setting for editorconfig

PR #981, opened 2021-04-20
Feature/parse references

JJK96:
Parse bible references in multiple languages using https://github.com/openbibleinfo/Bible-Passage-Reference-Parser.

It uses a fall through system, so the first language that successfully parses the input is used.
If the input cannot be parsed, the user has to try again.
OSIS references always parse correctly.

I tested it with English and Dutch references.

It successfully parses references like the following:
* John 3:16
* John.3.16
* Johannes 3:16
* Joh 3:16

The output is a link with as the text the entered text and as the OSIS link the result of the parsing.


JJK96:
Drawback of this PR is that the size of the .apk increases from 6.7M to 7.9M.
I think that is acceptable, but maybe others have a different opinion?

timbze:
> Drawback of this PR is that the size of the .apk increases from 6.7M to 7.9M.
> I think that is acceptable, but maybe others have a different opinion?

If we have significant benefits then I don't think it really matters. APK under 10Mb is still very good compared to most other Bible apps.

I think this is a great feature, though it would be even nicer, IMO, if it would automatically parse the note on save... 1. I don't know all the ins and outs of this, maybe that would not be easy to do. 2. it could be done at some point later as well.

Also, it would be nice if those references were stored in an index somewhere so that in the future we could easily find notes that contain a certain reference. That could also be a future project if you don't want to do it now.

JJK96:
> I think this is a great feature, though it would be even nicer, IMO, if it would automatically parse the note on save... 1. I don't know all the ins and outs of this, maybe that would not be easy to do. 2. it could be done at some point later as well.

I thought about this, I know that MySword does it like this, but it sometimes has it wrong and thinks that something is a reference while it isn't. So for now I think it's good to make it explicit that we want a reference. In the future it's easy to add automatic parsing, since the parsing library supports such a feature already. I think that such a feature should be in another PR though.

> Also, it would be nice if those references were stored in an index somewhere so that in the future we could easily find notes that contain a certain reference. That could also be a future project if you don't want to do it now.

Definitely a nice idea, but also definitely worth another PR.

tuomas2:
we can implement this as an add-on to keep size down. ill try this tomorrow.

JJK96:
If you want it as an add-on I'd recommend to have the english version available by default and the other languages as an add-on.

app/bibleview-js/src/components/TextEditor.vue
```diff
@@ -103,18 +103,50 @@ export default {
       title: 'Insert bible reference',
       result: async () => {
         const originalRange = document.getSelection().getRangeAt(0);
+        //Get the active languages and create a bible reference parser for each language
+        const languages = android.getActiveLanguages()
```
tuomas2: We should cache this. TextEditors are used extensively and querying this from jvm side every time does not make sense.

app/bibleview-js/src/components/TextEditor.vue
```diff
@@ -103,18 +103,50 @@ export default {
...
+        languages.forEach(async (lang) => {
+            try {
+                const bcv_parser = await import(`bible-passage-reference-parser/js/${lang}_bcv_parser`)
+                    .then((module) => {
+                        return new module.bcv_parser
+                    });
+                parsers.push(bcv_parser)
+            } catch (error) {
+                console.log(`No parser exists for language: ${lang}`)
+            }
```
tuomas2: also this importing I think we should do only once and use cached parsers from then on.

tuomas2: If module language is available (for all new bookmarks it is), I think we could try to use module language and UI language instead of trying everything?

app/bibleview-js/src/utils.js
```diff
@@ -40,8 +40,15 @@ export function setupElementEventListener(elementRef, eventType, handler, option
...
-            console.log(`Stub for ${key}(${args}) called`)
+        //Implement a separate stub for getActiveLanguages, since it needs to return data
+        if (key == "getActiveLanguages") {
+            stubs[key] = () => {
+                return ["he","nl","en"]
+            }
+        } else {
+            stubs[key] = (...args) => {
+                console.log(`Stub for ${key}(${args}) called`)
+            }
```
tuomas2: Instead of hard-coding here, I would rather like change signature to `stubsFor(object, defaults)`, such that we could
```
getStubsFor(exposed, {getActiveLanguages: ['he', nl', 'en']})
```
in android.js

app/bibleview-js/src/components/TextEditor.vue
JJK96: I tried to get the module language, but I couldn't find out how. Maybe you can shed some light on this.

app/bibleview-js/src/utils.js
JJK96: Nice idea

JJK96:
Thank you for the review, I'll look at caching and the stubs function.

app/bibleview-js/src/components/TextEditor.vue
tuomas2: If using bookmark: add language info to ClientBookmark from android side and pass it to TextEditor component via props or via provide()/inject(). 
If using document (for bibleview), add info to OsisFragment. (other options OsisDocument / BibleDocument) but I believe OsisFragment is good place; then it can be read on client side by injecting osisFragment).

JJK96: The module language should only take priority over the others, not replace it. Since I might want to make a Dutch comment on the KJV bible, and include a Dutch reference.

JJK96:
I updated the parser loading to use #987 and I added caching for the languages

JJK96:
TODO:

- [x] stubsFor
- [ ] ~~module language~~

app/bibleview-js/src/components/TextEditor.vue
```diff
@@ -51,9 +52,18 @@ export default {
   emits: ['save', "close"],
   setup(props, {emit}) {
     const android = inject("android");
+    const features = inject('customFeatures')
     const editorElement = ref(null);
     const editor = ref(null);
     const inputText = ref(null);
+    var languages = null
+
+    function get_languages() {
```
tuomas2: camelCaps

```suggestion
    function getLanguages() {
```

app/bibleview-js/src/components/TextEditor.vue
```diff
@@ -103,18 +113,50 @@ export default {
       title: 'Insert bible reference',
       result: async () => {
         const originalRange = document.getSelection().getRangeAt(0);
+        //Always add the "en" language, so at least the english parser always exists.
+        var en_parser = new bcv_parser;
+        var parsers = [en_parser]
+
+        if (features.features.refParser) {
+            //Get the active languages and create a bible reference parser for each language
+            const languages = get_languages()
```
tuomas2:
```suggestion
            const languages = getLanguages()
```

app/bibleview-js/src/components/TextEditor.vue
```diff
@@ -51,9 +52,18 @@ export default {
   emits: ['save', "close"],
   setup(props, {emit}) {
     const android = inject("android");
+    const features = inject('customFeatures')
     const editorElement = ref(null);
     const editor = ref(null);
     const inputText = ref(null);
+    var languages = null
```
tuomas2:
```suggestion
    let languages = null
```
(let or const as of es6 javascript, not old-style vars)

app/bibleview-js/src/components/TextEditor.vue
```diff
@@ -29,6 +29,7 @@
 import {inject, onMounted, onUnmounted, watch} from "@vue/runtime-core";
 import {ref} from "@vue/reactivity";
 import {useCommon} from "@/composables";
+import {bcv_parser} from "@/composables/refparser.js";
```
tuomas2: Why not use english version straight from the package?

app/bibleview-js/src/components/TextEditor.vue
```diff
@@ -103,18 +113,50 @@ export default {
...
+            languages.forEach(async (lang) => {
+                if (lang != "en") {
+                    try {
+                        const bcv_parser = await features.features.refParser(lang)
+                        parsers.push(bcv_parser)
+                    } catch (error) {
+                        console.log(`No parser exists for language: ${lang}`)
+                    }
+                }
+            })
```
tuomas2: Indentation (2 spaces in js side)

app/bibleview-js/src/components/TextEditor.vue
```diff
@@ -103,18 +113,50 @@ export default {
...
         const originalRange = document.getSelection().getRangeAt(0);
+        //Always add the "en" language, so at least the english parser always exists.
+        var en_parser = new bcv_parser;
+        var parsers = [en_parser]
+
+        if (features.features.refParser) {
+            //Get the active languages and create a bible reference parser for each language
+            const languages = get_languages()
+            languages.forEach(async (lang) => {
+                if (lang != "en") {
```
tuomas2:
```suggestion
                if (lang !== "en") {
```

app/bibleview-js/src/components/TextEditor.vue
```diff
@@ -103,18 +113,50 @@ export default {
       title: 'Insert bible reference',
       result: async () => {
         const originalRange = document.getSelection().getRangeAt(0);
+        //Always add the "en" language, so at least the english parser always exists.
+        var en_parser = new bcv_parser;
+        var parsers = [en_parser]
```
tuomas2:
```suggestion
        const en_parser = new bcv_parser;
        const parsers = [en_parser]
```
My ide (webstorm) suggests these changes automatically. Which IDE are you using?

app/bibleview-js/src/components/TextEditor.vue
JJK96: Because I think the English version is essential while the other languages should be optional. I didn't want to import from node_modules since I assumed that that would mean all are included.

app/bibleview-js/src/components/TextEditor.vue
JJK96: I'm not using an IDE, just a text editor. This is a drawback but it has other advantages.

app/bibleview-js/src/components/TextEditor.vue
```diff
@@ -103,18 +113,50 @@ export default {
...
+
+        if (features.features.refParser) {
+            //Get the active languages and create a bible reference parser for each language
+            const languages = get_languages()
+            languages.forEach(async (lang) => {
```
tuomas2: I guess we want to wait for this to finish?

```suggestion
            await Promise.all(languages.map(async (lang) => {
```

app/bibleview-js/src/components/TextEditor.vue
tuomas2: if we import only english, only english will be in the js bundle

tuomas2:
I moved parse logic to useParsers. Though somehow it does work with finnish at the moment

In finnish I tried "1. Moos 1:1" (which is Gen 1:1). Also tried Johannes 3:16 as it's Johannes in fi too.

Checked with your original version and finnish does not seem to work with that either, so it's probably not issue with my refactoring

app/bibleview-js/src/components/TextEditor.vue
```diff
@@ -103,18 +105,31 @@ export default {
...
+          //Keep trying to get a bible reference until either
+          //    * It is successfully parsed (parsed != "") or
+          //    * The user cancels (text === null)
+          while (parsed === "" && text !== null) {
+            text = await inputText.value.inputText();
+            if (text !== null) {
+              parsed = parse(text)
+            }
+            // TODO: give error message if it was not possible to parse.
+            // Use current value as initial text in input text.
```
tuomas2: - [x] this

tuomas2:
@JJK96 could you check if my modifications work for you (that I did not break nl references), and continue from here?

JJK96:
Screenshot_20210513-123806

@tuomas2 This reference works, so finish support is actually working. However "1. Moos 1:1" does not work, so it's probably a limitation of the parser.

tuomas2:
- [ ] we need to somehow instruct user to download refparser package, perhaps giving a button/link straight to download screen.

JJK96:
Making it a recommended module would be a good start. Or make an exclamation mark in the enter reference view with information, similar to the info for the editor.

app/bibleview-js/src/components/TextEditor.vue
JJK96: I'll skip this for now, using all parsers does not have any performance impact afaict.
In addition, getting the module language is tricky and it would require that we load parsers dynamically instead of only once.

agrogers:
I had problems with this recently. I thought I typed it correctly. 

image

JJK96:
@agrogers could you show the languages used for parsing and the href of the link?

tuomas2:
@agrogers this feature is not yet merged and thus not included in beta builds.

agrogers:
> @agrogers could you show the languages used for parsing and the href of the link?

Happy to... but I don't know how to do that. I created the link via the bible reference button.

timbze:
@JJK96 just thought you might want to check these forks of the reference parser:
https://github.com/renehamburger/Bible-Passage-Reference-Parser (improvements in German and other repo maintenance improvements)
https://github.com/alerque/Bible-Passage-Reference-Parser (improvements in Turkish)

tuomas2:
> @JJK96 just thought you might want to check these forks of the reference parser:
> https://github.com/renehamburger/Bible-Passage-Reference-Parser (improvements in German and other repo maintenance improvements)
> https://github.com/alerque/Bible-Passage-Reference-Parser (improvements in Turkish)

I can include improved versions in RefParser package

tuomas2:
Added couple of strings in dev to support help text 0fcf3cc6c42a919b3633104a883edcca94364fd3

JJK96:
I have added the help functionality:

reference_modal
reference_modal_help

agrogers:
Thanks for your work on this. Should this be working in English in the beta
Tuomas has published? I still just get a prompt to "Please download 'John
1'"
It's never worked for me. Actually, initially I thought it was a pointless
feature but there have been quite a few occasions where i have wanted it. A
verse selector will be good - or even just a book selector - I never sure
exactly what format i should enter this as .

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

tuomas2:
> Thanks for your work on this. Should this be working in English in the beta Tuomas has published? I still just get a prompt to "Please download 'John 1'" It's never worked for me. Actually, initially I thought it was a pointless feature but there have been quite a few occasions where i have wanted it. A verse selector will be good - or even just a book selector - I never sure exactly what format i should enter this as .

No it's not yet included. You will see if it is included or not by following this pull request status. When it's "open" or "closed", it's generally not included, when it's "merged" it's included.

agrogers:
Thanks!

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

tuomas2:
@JJK96  pushed some style improvements to your branch

tuomas2:
Bug: After RefParser module is installed, we need to restart app to see it. (will create ticket)

tuomas2:
Thank you @JJK96 for this feature!

agrogers:
It worked for me for the first time. Thanks!

PR #983, opened 2021-04-21
Fix accented character "a" for speak

### LINKED ISSUE #982: User-definable list of string replacements for TTS
We need to be able to define list of replacements in (advanced?) Speak settings. TTS engines behave differently so we don't want to have any hard-coded strings in general.

Perhaps also supporting regex for advanced users?

Original:
> **Describe the bug**
> Speak says "a con asento" (a with accent) for the letter "á" when it stands alone
> 
> **Bug was found on And Bible version**
> Build 3.3.399
> 
> **To Reproduce**
> Speak Deut 20:1 in RV1909 spanish module (third word)

timbze:
Closes #982 

I don't know if this is a proper fix, but it works. Any suggestions @tuomas2?

### DIFF
--- a/app/src/main/java/net/bible/service/format/osistohtml/osishandlers/OsisToBibleSpeak.kt
+++ b/app/src/main/java/net/bible/service/format/osistohtml/osishandlers/OsisToBibleSpeak.kt
@@ -152,6 +152,7 @@ class OsisToBibleSpeak(val speakSettings: SpeakSettings, val language: String) :
         s = s.replace("`", "'")
         s = s.replace("´", "'")
         s = s.replace("’", "'")
+        s = handleAAccent(s)
         s = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
             Html.fromHtml(s, 0).toString()
         }
@@ -178,6 +179,17 @@ class OsisToBibleSpeak(val speakSettings: SpeakSettings, val language: String) :
                 speakCommands.add(TextCommand(s))
             }
         }
+    }
+
+    /**
+     * Speech in spanish says "a con asento" (a with accent). Letter is also used in other languages.
+     * Only replace when letter stands alone, never replace in a word.
+     */
+    private fun handleAAccent(s: String): String {
+        val firstLetter = Regex("^[Áá](?:[\\s])")
+        val lastLetter = Regex("(?:[\\s])[Áá]$")
+        val wordAlone = Regex("(?:[\\s])[Áá](?:[\\s])")
+        return s.replace(firstLetter, "a").replace(lastLetter, "a").replace(wordAlone, "a")
     }
 }
 


app/src/main/java/net/bible/service/format/osistohtml/osishandlers/OsisToBibleSpeak.kt
```diff
@@ -179,5 +180,16 @@ class OsisToBibleSpeak(val speakSettings: SpeakSettings, val language: String) :
...
+    /**
+     * Speech in spanish says "a con asento" (a with accent). Letter is also used in other languages.
+     * Only replace when letter stands alone, never replace in a word.
+     */
+    private fun handleAAccent(s: String): String {
+        val firstLetter = Regex("^[Áá](?:[\\s])")
+        val lastLetter = Regex("(?:[\\s])[Áá]$")
+        val wordAlone = Regex("(?:[\\s])[Áá](?:[\\s])")
+        return s.replace(firstLetter, "a").replace(lastLetter, "a").replace(wordAlone, "a")
+    }
```
tuomas2, line 193: Essentially we are fixing spanish text to speech synthesis engine behavior? If this is spanish-only issue should we make this happen only to spahish text (check doc language)?

I don't really like "patching" text to speech engine behavior too much on our side. Google TTS has it's weaknesses (I'm also suffering from the fact that there are quite many words in finnish that will be pronounced english way, but there's nothing I can do about it).

timbze: There are quite a few languages that use that type of letter, but I have not checked how Google TTS handles those languages.

If it was just pronunciation I wouldn't care that much, but this is a bit worse. If you'd say "there's á tree in the yard", TTS would say "there's a with accent tree in the yard".

Maybe I should check some of [these other languages](https://en.wikipedia.org/wiki/%C3%81) first to get a better idea? (i.e. Chinese (Pinyin), Blackfoot, Czech, Dutch, Faroese, Galician, Hungarian, Icelandic, Irish, Kazakh, Lakota, Navajo, Occitan, Portuguese, Sámi, Slovak, Spanish, Vietnamese, Welsh, and Western Apache) 

Not sure if I'd catch it though for languages that are totally strange for me.

tuomas2: I see. Are these accented letters in all spanish texts? What about spanish AB modules?
I'm wondering if we should do some user settings (even user-configurable search&replace list) in advanced tts settings. Or some options that certain fixes can be enabled/disabled, as it's highly depending on TTS engine which is being used.

timbze: > Are these accented letters in all spanish texts?

No, it seems that not all modules use it. The Reina Valera 1909 from both eBible and Crosswire have it, but some other versions I checked do not.

I think you're right, it would probably be better to make a setting like this optional

timbze:
> Tuomas said: I'm wondering if we should do some user settings (even user-configurable search&replace list) in advanced tts settings. Or some options that certain fixes can be enabled/disabled, as it's highly depending on TTS engine which is being used.

I agree that should be done and since I won't be able to spend enough time to do that now I will close the PR for now

PR #984, opened 2021-04-21
Remove parenthesis "()" from speech text to prevent TTS saying sad face etc

### LINKED ISSUE #356: Speak says "Sad Face"
When using the speak feature, AB reads "sad face", probably where there should be a space in between verses.

In the below verses between 15 and 16, AB speak reads: "the Holy Ghost sad face for as yet"

(Acts 8:15-16)
 Who, when they were come down, prayed for them, that they might receive **the Holy Ghost:  16.(For as yet** he was fallen upon none of them: only they were baptized in the name of the Lord Jesus.) 

This is the text TTS outputs
image

**Smartphone**
 - Device: LG G6
 - Version Android 8.0

timbze:
Closes #356

### DIFF
--- a/app/src/main/java/net/bible/service/format/osistohtml/osishandlers/OsisToBibleSpeak.kt
+++ b/app/src/main/java/net/bible/service/format/osistohtml/osishandlers/OsisToBibleSpeak.kt
@@ -152,6 +152,8 @@ class OsisToBibleSpeak(val speakSettings: SpeakSettings, val language: String) :
         s = s.replace("`", "'")
         s = s.replace("´", "'")
         s = s.replace("’", "'")
+        s = s.replace("(", "")
+        s = s.replace(")", "")
         s = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
             Html.fromHtml(s, 0).toString()
         }


app/src/main/java/net/bible/service/format/osistohtml/osishandlers/OsisToBibleSpeak.kt
```diff
@@ -152,6 +152,8 @@ class OsisToBibleSpeak(val speakSettings: SpeakSettings, val language: String) :
         s = s.replace("`", "'")
         s = s.replace("´", "'")
         s = s.replace("’", "'")
+        s = s.replace("(", "")
+        s = s.replace(")", "")
```
tuomas2, line 156: Should we replace it with space instead? "Test(test two)" would render as "Test test two". Though then "Something (test). Something" would render as "Something test . Something" (period would render alone). 

Should we just replace sad faces (":(" -> ":")?

timbze: Replacing with space would probably work. I tried adding space before and after parenthesis but that did not work, I'm guessing later step may remove space again.

I've also heard TTS say wink-face and smile-face 😄 so we'd have to replace those too if we wanted to go that route...

Perhaps add a . (dot) between :) ;) :( ;( so they would become :.) ;.) etc...
-- however, I find that doesn't work at this level because I in Acts 8:15-16 the colon is separate (alone) from the parenthesis...

timbze:
> Tuomas said in #983: I'm wondering if we should do some user settings (even user-configurable search&replace list) in advanced tts settings. Or some options that certain fixes can be enabled/disabled, as it's highly depending on TTS engine which is being used.

I agree that should be done and since I won't be able to spend enough time to do that now I will close the PR for now

PR #987, opened 2021-04-22
Implement secure support for external js files

tuomas2:


tuomas2:
@JJK96 this would be basis for using RefParser from module zip. I created RefParser.zip too, I'll upload it to beta repository soon..

tuomas2:
Added RefParser Add-on module to beta repository.

Issue #332, opened 2019-02-27
Bible book chooser improvements


tuomas2:
- Remove gradient (check if it looks good)
- Transpose books #142 
- High contrast indicator for currently selected book in book chooser menu #97
- Ideas #715
- Optional **quick one-view chooser**(s), similar to what is in MySword (see screenshot)

![Screenshot_20210323-124052](https://user-images.githubusercontent.com/5811789/112134218-2af61f80-8bd5-11eb-80a2-18dbad5b1e55.png)

tuomas2:
There was typo in commit msg.. (4a8771c)

tuomas2:
From darryl:

>  Please consider indicating the current book, in the Book Selector, in INVERSE colours. Eg: when Matthew (usually Orange text on black gradient) becomes current, then it should be Black text on Orange background; Romans would be Black text on yellow background (when current); etc.

tuomas2:
User via Google play:
> I love this bible so much, although I will shoot you mails, I am having problems with the choose book interface. Could you give us options? Maybe the old testament scrollable on one side and the new testament scrollable on the other. Then without the colours, just on a transparent white background to ease quick recognition?

Screwtapello:
Currently, the Bible book chooser shows the abbreviation for each book (Job, Joe, Joh, Jon, etc.) and when you press a particular button, a tooltip appears above the button so you can see which button is obscured by your finger.

This is great, but I think it would be even better if the tooltip showed the long name of the book, as confirmation I'm pressing the correct button. I understand the buttons themselves have to show the abbreviation due to space constraints, but the tooltip can take much more space, and some of those abbreviations are confusingly terse.

- [ ] Showing the full name of the book would mean a new Christian has one less thing they need to memorise in order to find their way around the Bible.

cmahte:
Including both the interface abbreviation and a title from within the text currently displayed adds value beyond "new christian." 

Abbr (TITLE in the Current work) would be perfect.  not only display the available interface translation of the abbr, but a title in the language of the text currently displayed. So not just a 'newbie' feature. And for folks like me... helps to confirm the physical book i'm comparing a digital reference to... that  I've interpreted the header text on the physical page correctly in choosing my lookup. 

As opposed to having to look in chapter 1 first for the language  of the title match... 

If this opens up for edit.... just a thought.

timbze:
From someone with sight impairment:

[...]

Issue #142, opened 2018-07-15
Transpose books of the bible grid to read from left to right.


steelwil:
I think this would be easier to read. I find it difficult reading from top to bottom.

cmahte:
I think the current order is fine, but I see value in providing the user
the ability to change the sorting, especially to alphabetical. New
believers struggle with canonical order, especially in areas that have a
mix of bible orders present.

How about a setting. A global option button to choose the sort order for
verse chooser. I don't know if it should also affect scrolling (to next
book) or not, but that might be worth considering.

Field 1: Sort Horizontal (button with an right arrow), Sort Vertical
(button with a down arrow)
Field 2: Sort a Dropdown list: (Per module versification, Protestant,
Catholic, Orthodox, Jewish, Alphabetical, Chronological. etc.)

oxr463:
## Responsive interface

Instead of showing a grid, is it possible to have each book on its own line?

**Current**

| | | | | | |
|-|-|-|-|-|-|
| Gen | 2Ki | Isa | Nah | Rom | Tit |
| Exo | ... | ... | ... | ... | ... |

**Desired**

| |
| - |
| Genesis |
| Exodus |
| ... |

--

This would be helpful for smaller screens. I am always accidentally pressing the wrong book on my phone, (Tracfone Alcatel Pixi Eclipse w/ 4.0" display).

Resources: https://github.com/googlesamples/android-ConstraintLayoutExamples

oxr463:
Here is the original `choose_passage_book_menu` file:

```
<?xml version="1.0" encoding="utf-8"?>
<!-- The main menu accessed via Menu key

 @author Martin Denham [mjdenham at gmail dot com]
-->
<menu xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:appcompat="http://schemas.android.com/apk/res-auto" >


</menu>
```

Which produces the following result:

![Choose Book](https://i.stack.imgur.com/PeP9P.png)

And here is the new one with changes applied to it:

```
<?xml version="1.0" encoding="utf-8"?>
<!-- The main menu accessed via Menu key

 @author Martin Denham [mjdenham at gmail dot com]
-->
<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:appcompat="http://schemas.android.com/apk/res-auto"
    android:layout_width="match_parent"
    android:layout_height="match_parent"
    android:orientation="vertical">
<menu></menu>
</ScrollView>
```

[...]

PR #988, opened 2021-04-22
Highlight currently selected book in book chooser menu

jmarsican:
**Describe the pull request content**
- One liner solution to set as pressed the button corresponding to highlighted in when choosing passage
- Previously selected book and chapter will be highlighted to ease jumping to  another passage and give a sense of  orientation so user experience will be  improved for passage selection while  navgating through  several passages in the Bible
- Fixes issue #97 

**Screenshots**
book
chptr
landscape

PR #994, opened 2021-04-25
Book and chapter selector UX improvements

jmarsican:
**Book and chapter selector user experience improvements for easy and fast navigation among passages**
- Use modern look & feel for buttons with default material design instead of gradient
- Show full book name hint
- Transpose books buttons table grid
- Use slightly different color for new testament buttons to reduce visual effort while navigating among passages
- Improve book preview hint text
- Tweaks and fixes for #332 #142 #715 

**Screenshots**

finalLands

finalBook


tuomas2:
Looks good! Checking this more carefully on next thursday.

timbze:
Very nice UI improvements! I have a few suggestions, but @tuomas2 should affirm or deny my suggestions before you make changes based on them.

- [ ] Leave the default book order the way it was (in portrait mode the books go by column, in landscape by row)
- [x] Add back (up) button in toolbar (can be done by modifying manifest file) #997
- [ ] All settings would save in Db that way the user's preference would again be used next time
- [ ] Instead of current toolbar sort button on right, add a 3-dot menu with:
  * Sort by book order [checkbox in menu group 1] << this would be default
  * Sort alphabetically [checkbox in menu group 1]
  * Separator
  * Order books by row [checkbox in menu group 2] << off by default, but when checked the books would also order by row in portrait mode the way landscape already does by default

tuomas2:
Good suggestions @timbze. Settings should be to shared preferences (CommonUtils.sharedPreferences).

tuomas2:
@timbze problem with "back" button is that this activity is used from many places (not just mainbibleactivity, but at least from SpeakActivity as well)

timbze:
> @timbze problem with "back" button is that this activity is used from many places (not just mainbibleactivity, but at least from SpeakActivity as well)

Ok in that case it would need to be manually implemented to run "back" action right? I think I've done it with a few lines of code.

I just think the screen would look more user friendly if it had the back button

tuomas2:
> > @timbze problem with "back" button is that this activity is used from many places (not just mainbibleactivity, but at least from SpeakActivity as well)
> 
> Ok in that case it would need to be manually implemented to run "back" action right? I think I've done it with a few lines of code.

Code example for this would be nice. I'm in the impression that it's not very convenient to override default behavior. (Kind of remember trying this at some point)

> I just think the screen would look more user friendly if it had the back button

I agree.

timbze:
EDIT: Made PR #997

> Code example for this would be nice.

See https://github.com/timbze/and-bible/tree/demo/implement_back_action
* Code back (up) button for GridChoosePassageBook: https://github.com/timbze/and-bible/commit/0af9bd9b2a3d89b0f709101881b072db9bf627da
* Manifest.xml back (up) for GridChoosePassageChapter and Verse https://github.com/timbze/and-bible/commit/1f82066b329e0843d494ae0c32b70054e71305c6

@tuomas2 if you see something wrong with this or something that might not work let me know. I have used this type of code in my personal apps.

jmarsican:
Glad you liked these improvements folks. The purpose of this PR was to address items specified in #332 as per @tuomas2 requested in my previous PR #988.  I like suggestions @timbze made too, but perhaps they are out of scope and might be addressed in another PR. 
I can revert changes to traspose books grid order and would try to investigate how to implement such 3-dot menu button with those three sort/ordering strategies. Although, it might take long time to understand and design implementation (I don't know how local storage might be implemented) and I will not be able to do it before next weekend.

timbze:
@jmarsican thanks for your help in contributing to the app, we're always glad for contributors!

Pull requests are usually not a rush, we work on stuff as we have time.

I can give you pointers on how to implement the 3-dot menu and saving of settings, it's really not that hard since most hard work in the code is already there.

I wouldn't mind if the other things were in a separate PR, but the default order needs to stay the way it was, so if you revert that then feature-wise IMO this PR would be ready to merge.

I have at least one code suggestion. Will look over it again and submit code reviews if I have any other.

app/src/main/java/net/bible/android/view/activity/navigation/GridChoosePassageBook.kt
```diff
@@ -47,12 +47,14 @@ import java.util.ArrayList
 
 import javax.inject.Inject
 
+private val NEW_TESTAMENT_TINT = Color.parseColor("#FF505050")
```
timbze: Variables should generally be in a class. My suggestion is to move this to the area where `isCurrentlyShowingScripture` is.

app/src/main/java/net/bible/android/view/activity/navigation/GridChoosePassageChapter.kt
```diff
@@ -103,6 +103,7 @@ class GridChoosePassageChapter : CustomTitlebarActivityBase(), OnButtonGridActio
             // this is used for preview
             buttonInfo.id = i
             buttonInfo.name = Integer.toString(i)
+            buttonInfo.description = Integer.toString(i)
```
timbze: Should be `i.toString()`. Also the line above it.

jmarsican:
> @jmarsican thanks for your help in contributing to the app, we're always glad for contributors!
> 
> Pull requests are usually not a rush, we work on stuff as we have time.
> 
> I can give you pointers on how to implement the 3-dot menu and saving of settings, it's really not that hard since most hard work in the code is already there.
> 
> I wouldn't mind if the other things were in a separate PR, but the default order needs to stay the way it was, so if you revert that then feature-wise IMO this PR would be ready to merge.
> 
> I have at least one code suggestion. Will look over it again and submit code reviews if I have any other.

For sure Timmy. I am going to restore default order the way it was shortly. 

Please do share any clue on how to implement saving settings so I can analyze and implement as soon as I can.

Looking forward to get any code suggestions that might improve the code of this PR.

Best Regards.

tuomas2:
Most recent example for saving sharedPreferences:
ShareWidget.kt 
reading on line 44
saving on line 65

app/src/main/java/net/bible/android/view/util/buttongrid/ButtonGrid.kt
```diff
@@ -218,14 +212,13 @@ class ButtonGrid constructor(context: Context, attrs: AttributeSet? = null, defS
 
     private fun showPreview(buttonInfo: ButtonInfo) {
         try {
-            val button = buttonInfo.button
             if (buttonInfo != currentPreview) {
-                Log.d(TAG, "Previewing " + buttonInfo.name)
+                Log.d(TAG, "Previewing " + buttonInfo.description)
                 currentPreview = buttonInfo
-                previewText!!.text = buttonInfo.name
+                previewText!!.text = buttonInfo.description
```
timbze: The double exclamation `!!` should not be used normally. It causes a crash when there's null, and especially in cases like this where it's not some important app logic, it should never cause a crash (even though it should never be null). I see you just used what was there from before and that's not a problem 😄 

It would be better to add a new line under `Log.d` (line 216), like this:
`val preview = previewText ?: return`

And then now use `preview.` everywhere underneath where there previously was `previewText!!.`

app/src/main/java/net/bible/android/view/util/buttongrid/ButtonGrid.kt
```diff
@@ -218,14 +212,13 @@ class ButtonGrid constructor(context: Context, attrs: AttributeSet? = null, defS
...
                 val popupHeight = previewHeight
                 previewText!!.measure(MeasureSpec.makeMeasureSpec(0, MeasureSpec.UNSPECIFIED), MeasureSpec.makeMeasureSpec(0, MeasureSpec.UNSPECIFIED))
-                val popupWidth = Math.max(previewText!!.measuredWidth, buttonInfo.button!!.width + previewText!!.paddingLeft + previewText!!.paddingRight)
+                val popupWidth = max(previewText!!.measuredWidth, buttonInfo.button!!.width + previewText!!.paddingLeft + previewText!!.paddingRight)
```
timbze: Android Studio shows the `!!` after `.button` is useless, you can remove that as well

app/src/main/java/net/bible/android/view/util/buttongrid/ButtonGrid.kt
jmarsican: Yes my bad. I use to follow boy scout rule but I forgot for this class. Looks like it was a Java class automatically converted to Kotlin and there are plenty of Java-ish remains to be reviewes :sweat_smile:

jmarsican:
@timbze all code improvements you suggested were resolved.
@tuomas2 I've just restored default book order the way it was. Also I reported in #1013 feature to allow user chose preferred order, I'll work on it next weekend.

tuomas2 (approved):
Looks good, thank you! :-)

PR #1004, opened 2021-05-01
Instead of crashing if speak could not be toggled, show a toast instead.

RuggMatt:
If you try to currently toggle speak while on a study pad, the app ends up crashing

I couldn't figure out how to disable the speak button for study pads, as it is categorized as a General Book, so instead of invoking a crash, perhaps it would be better to just show a Toast error instead.


tuomas2:
there is some method isspeakable or something in page classes

RuggMatt:
> there is some method isspeakable or something in page classes

Yes I found that but it is set per book category. Study pads are categorized as general books. Other general books seem to work fine when using TTS, but study pads don't work. I could hard code the book name, since it shouldn't change,  but looking for some other way that isn't as hard coded.

PR #968, opened 2021-04-11
Feature - Import reading plan

timbze:
Closes #180
Closes #570
Closes #938
Closes #918 (4818070d)

- [ ] Reading plan import validation for #180

PR #1011, opened 2021-05-02
Reading plan import strings

timbze:
I think this is the strings that will be needed for #968


timbze:
For my part this PR is ready to merge. 

I wasn't sure if this string needed translation help (the %s would have medium app name). But don't think it's really necessary. `<string name="reading_plan_ships_with_ab">This reading plan ships with %s</string>`

Issue #680, opened 2020-05-27
Store application preferences in database


tuomas2:
instead of 'shared preferences'

timbze:
I think it would be nice to find (or build) a little library to be able to use simple commands to write and get prefs to database similar to how shared preferences works...

tuomas2:
We just need to implement our own PreferenceDataStore and use it to read/write from db. Similarly as ColorSettingsDataStore and TextDisplaySettingsDataStore.

tuomas2:
Got accross this: https://developer.android.com/topic/libraries/architecture/datastore
Not sure if this is optimal way to go either.

tuomas2:
I believe better to implement this manually by making couple of utility functions that access db.

bvahedy:
I am with Toumas on this one, because if you share your database with
someone then their preferences will be over written.

agrogers:
Hadn't thought about sharing database without the preferences. Could the
restoration process just give an option to exclude application preferences?

I dont think I have ever shared my database with someone so i would say the
default should be to restore those preferences.

Also, my database and my documents are quite tightly connected. I have
workspaces for particular documents.

___________________________
*www.TenOutOfTen.org* <https://www.TenOutOfTen.org>
***@***.***
(+61) 0477 174 193 (Australia)
(+95) 09 250018669 (Myanmar)

bvahedy:
At the risk of not being labled as a male chauvinist pig, many of our wives
(no, I don't mean having multiple wives), if they are spiritual hungry
enough, would like to have on their devices our bookmarks and lables for
topical studies, or to learn their husband's insights of the scriptures.
since women technically are not spiritually that savvy. (Ok I've done it
now 😁)
Ok seriously, the app settings doesn't have to be in the database, it could
be  in an ini file in the backup folder on the device and if upon the
initial install that file exists then the app uses that file to set all the
app settings, otherwise it sets everything up just as it does now.
But we would have to add in the backup menu an item to save the app
settings, and that would create or update the ini file.

I would like to know what Toumas' thoughts are on this.

agrogers:
I would much prefer to know your wife's thoughts on that email :)

An ini file would be fine for me. That probably wouldn't help with a new
device though. I have noticed it because i have done a lot of reinstalling
recently. Redoing those settings every few years is probably not a bother.

___________________________
*www.TenOutOfTen.org* <https://www.TenOutOfTen.org>
***@***.***
(+61) 0477 174 193 (Australia)
[...]

PR #1055, opened 2021-05-14
Start working on custom store to app prefs

tuomas2:
#680 

It looks like it's not quite possible to set custom store to application-wide SharedPreferences, but it is via settings activity.


tuomas2:
Implemented other way in #1308

PR #1079, opened 2021-05-20
Previously existing bookmark highlight range adjustment

### LINKED ISSUE #1078: Adjust bookmark's highlight range afterwards

tuomas2:
Closes #1078

### DIFF
--- a/app/bibleview-js/src/components/BookmarkButtons.vue
+++ b/app/bibleview-js/src/components/BookmarkButtons.vue
@@ -30,6 +30,9 @@
       <div :class="{highlighted: !bookmark.wholeVerse}" class="bookmark-button" @click="toggleWholeVerse">
         <FontAwesomeIcon icon="text-width"/>
       </div>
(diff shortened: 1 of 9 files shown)


tuomas2:
Can't figure out how to trigger floating action mode (which also makes range adjustment handles visible) from code.

tuomas2:
Related question (but no answer unfortunately): https://stackoverflow.com/questions/60540521/how-to-show-selection-cursor-after-change-the-range-of-selection-programmaticall

agrogers:
What if you did it this way:
1. Select a range using the long click method
2. Add an additional option to the pop up menu called 'Change Bookmark'
3. Throw up a dialog that lists all the bookmarks that the range touches.
4. Let the user choose the bookmark they want to adjust and simply update the range for that bookmark based on the new selection.
5. The dialog might need an option to specify if the selection applies to just the selected text or to the verses the selected text covers. You could add that as another menu option in the popup menu but that menu is not beautiful and adding more options would make it even less beautiful.

PR #1118, opened 2021-05-31
Show the "Up" button on WorkspaceSelectorActivity

### LINKED ISSUE #1111: Add back (up) button to "Select workspace" window

uniqueid001:
Resolves #1111

### DIFF
--- a/app/src/main/java/net/bible/android/view/activity/workspaces/WorkspaceSelectorActivity.kt
+++ b/app/src/main/java/net/bible/android/view/activity/workspaces/WorkspaceSelectorActivity.kt
@@ -164,6 +164,7 @@ class WorkspaceSelectorActivity: ActivityBase() {
         resultIntent = Intent(this, this::class.java)
         binding = WorkspaceSelectorBinding.inflate(layoutInflater)
         setContentView(binding.root)
+        this.supportActionBar?.setDisplayHomeAsUpEnabled(true)
 
         workspaceAdapter = WorkspaceAdapter(this).apply {
             setHasStableIds(true)
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -151,6 +151,7 @@
             android:name="net.bible.android.view.activity.workspaces.WorkspaceSelectorActivity"
             android:configChanges="keyboardHidden|orientation|screenSize|locale"
             android:label="@string/workspace_selector_title"
+            android:parentActivityName="net.bible.android.view.activity.page.MainBibleActivity"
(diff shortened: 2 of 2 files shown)


app/src/main/java/net/bible/android/view/activity/workspaces/WorkspaceSelectorActivity.kt
```diff
@@ -162,6 +162,7 @@ class WorkspaceSelectorActivity: ActivityBase() {
         windowControl.windowRepository.saveIntoDb()
         resultIntent = Intent(this, this::class.java)
         setContentView(R.layout.workspace_selector)
+        this.supportActionBar?.setDisplayHomeAsUpEnabled(true)
```
timbze, line 167: Is this line even necessary? I didn't test it here yet, but I didn't think this is necessary when adding the line in Manifest file.

uniqueid001: Good question.  I automatically added it, since that is my experience from work, although at the time we also needed the backwards-compatible meta-data element way of specifying the parent activity.  I did some more investigation and found that the button does display with just the attribute added to the Manifest.  It also appears to not honor the activity specified in the Manifest: no matter which activity I specified it went to the top-level MainBibleActivity, unless I specified nonsense and then it logged the error instead.  The documentation for that method indicates that setting it to true indicates that it will navigate up one level in the hierarchy and leaving it false will navigate to the top-level of the hierarchy.  So it may not be meaningful for this activity.

timbze: I think in this case we are always going back to main activity so we can just use the line in Manifest file.

In my experience, if setting the `setDisplayHomeAsUpEnabled()` method, you also have to implement the override `onOptionsItemSelected`. In that case the line in Manifest is not needed. I think otherwise it's useless. See [this commit](https://github.com/AndBible/and-bible/pull/997/commits/696a015955175da69778e8afabe0fcf31b495574) for example.

> The documentation for that method indicates that setting it to true indicates that it will navigate up one level in the hierarchy and leaving it false will navigate to the top-level of the hierarchy.

I think this is when you use `NavigationManager`. Setting to false makes it the home button, setting to true, the "up" button. If I'm right I think that is what it's referring to. When not using `NavigationManager`, it has to manually implemented like in my linked commit above.

tuomas2 (approved):

tuomas2:
Thank you for PR!

PR #1214, opened 2021-06-30
Apply the grid sorting preference to chapter & verse

### LINKED ISSUE #1208: When books are sorted horizontally, sort chapter/verse numbers horizontally as well
@jmarsican (and others) what do you think? I think it makes sense.

### LINKED ISSUE #1213: Provide row-major ordering for the Choose Chapter grid, and the Choose Verse grid
#1122 added a user-controlled setting for ordering the Bible book chooser in row-major order, however it leaves the chapter and verse grids in column-major order.  I would like to apply row-major order to the chapter, and verse, selection grids.

_Design questions:_  

1. Should each of the three choosers have a separate preference?  I think this is excessive: I expect that others who want row-major order in one of the choosers will also want it in all three.
2. If there is only a single preference, should the chapter and verse activities also provide a menu item for toggling it?

uniqueid001:
Apply the grid sorting preference for Bible books to the chapter and
verse selection grids.

Closes #1213 
Closes #1208


tuomas2 (approved):
Thank you!

Issue #1236, opened 2021-07-04
Restore strongs links formatting style to be like in 3.3


agrogers:
The old style which I quite liked looked like this...
![IMG_20210704_171419.jpg](https://user-images.githubusercontent.com/13920678/124379691-d0a04b00-dcfb-11eb-83eb-5d433452d360.jpg)

The new style that I see looks like this...
![IMG_20210704_191215.jpg](https://user-images.githubusercontent.com/13920678/124379697-ddbd3a00-dcfb-11eb-87e4-be2578375d76.jpg)

I preferred the old style because the numbers were much less intrusive.

tuomas2:
Need to re-implement original css in new bibleview.js (in corresponding vue component). Your 2nd PR? :-D

agrogers:
Sure :D  Nice to do some work rather than just making work for others!

[Maybe the below should be in another ticket]
Is this something that should be in an Add On? That is, All formatting like
this is in a separate file(s). Users could then (provide) and install
whatever format style they like.

I would also like to make the Strongs more beautiful (see below). Perhaps
that could be done through an addon config file. Our formatting of the
definitions is functional but not so beautiful.

[image: image.png]


___________________________
*www.TenOutOfTen.org* <https://www.TenOutOfTen.org>
***@***.***
(+61) 0477 174 193 (Australia)
(+95) 09 250018669 (Myanmar)

tuomas2:
> Sure :D Nice to do some work rather than just making work for others! [Maybe the below should be in another ticket] Is this something that should be in an Add On? That is, All formatting like this is in a separate file(s). Users could then (provide) and install whatever format style they like. I would also like to make the Strongs more beautiful (see below). Perhaps that could be done through an addon config file. Our formatting of the definitions is functional but not so beautiful. [image: image.png]

Image is not visible. (GH does not support email reply attachments)

Now what I'm suggesting is that you will edit the component directly (it's W.vue, see <template> and <style> blocks)).
You look for style.css from old code (3.3), how it was styled earlier. #1222 will be making it possible to create overriding css that will customize anything displayed in And Bible's WebView.

bvahedy:
Andrew, your image didn't come across because you used the email to reply,
please use GitHub to include the image.

agrogers:
This is the missing image. Do you think that could be accomplished with css only? 

[...]

PR #1281, opened 2021-07-09
Formatting of Strongs numbers

agrogers:
I couldn't test against the Morph codes because I couldn't work out how to get them to show on my install.

image

### DIFF
--- a/app/bibleview-js/src/components/BibleView.vue
+++ b/app/bibleview-js/src/components/BibleView.vue
@@ -417,9 +417,4 @@ a {
   text-decoration: none;
   display: inline-block;
 }
-.strongs {
-  font-size: 0.6em;
-  text-decoration: none;
-  color: coral;
-}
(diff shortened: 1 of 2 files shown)


app/bibleview-js/src/components/BibleView.vue
```diff
@@ -417,4 +417,9 @@ a {
   text-decoration: none;
   display: inline-block;
 }
+.strongs {
+  font-size: 0.6em;
+  text-decoration: none;
+  color: coral;
+}
```
tuomas2: These lines should be in W.vue

agrogers: Ok, I have moved those changes across. So now i think it is correct. Yes?

tuomas2 (approved):

tuomas2:
Closes #1236

PR #1292, opened 2021-07-10
Bible and New Window updated icons

agrogers:
Two new icons for you to check. Have I made the necessary changes to use these correctly?
It's good to see the icons in place. I can adjust or revert if needed.


tuomas2:
Let's try

app/src/main/res/drawable/ic_add_circle_outline_white_24dp.xml
```diff
@@ -14,6 +14,8 @@
   ~ You should have received a copy of the GNU General Public License along with And Bible.
   ~ If not, see http://www.gnu.org/licenses/.
   ~
+
+  File was used for New Window menu option.
```
tuomas2, line 18: if this is not  any more used, it should be removed

PR #1293, opened 2021-07-10
New icons and menu arrangement

agrogers:
I took some liberties and arranged the menus how I feel they should be ordered. For example, all workspace related settings should be together and underneath 'Workspace'. Actions that manipulate the window itself should be together. Maximise should be at the top. Close should be at the bottom, etc etc.  I should have done the icons first, then the menu rearrangement in another pull request cause now it is hard to revert to the old method without losing all of the changes. I'm happy to rearrange if needed.

The annotations below are things I would like to try. But i think there  is a real love for some of those menu options :)

image

image


tuomas2:
let's try!

PR #1332, opened 2021-07-17
Alignment, boxes and other formatting in the study pad

agrogers:
This was bothering my obsessive compulsive parts. Is this used elsewhere where the left margin is required?

Screenshot 2021-07-17 132458

Screenshot 2021-07-17 132516

### DIFF
--- a/app/bibleview-js/src/components/OSIS/W.vue
+++ b/app/bibleview-js/src/components/OSIS/W.vue
@@ -18,7 +18,7 @@
 <template>
   <template v-if="showStrongsSeparately">
     <template v-if="(showStrongs && lemma) && (config.showMorphology && morph)">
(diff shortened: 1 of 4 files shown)


app/bibleview-js/src/components/OSIS/W.vue
```diff
@@ -112,9 +112,13 @@ export default {
   .link-style {
     text-decoration: underline dotted;
   }
-.strongs {
+.skip-offset {
```
tuomas2, line 115: skip-offset should not be styled. It's markup for highlight code. Better add separate class.

tuomas2: I'll fix this after merge.

PR #1381, opened 2021-07-23
More icon additions and changes

agrogers:


tuomas2:
As it looks like you are still working on this, let's change it to "drawft" pull request. You can change status back when it is ready to be merged.

agrogers:
Ah, that's what 'draft' means. I will use that in the future. Thanks.

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

agrogers:
> Removed the margin and padding around the outside of the edit control. The original CSS applied to all modals so needed to get around that with an extra class property.
> 
> image

I tried to revert this edit box change in this pull request but left it in the original pull request. Maybe i have done the wrong thing. Sorry!

agrogers:
I have made a mess with this one. Reverted something i shouldn't  have I think. So I will close this request and see if i can make another that  is cleaner.

PR #1397, opened 2021-07-24
Show two color underline

agrogers:
This uses the border style to produce the second underline. The second color is determined by either the first or second label in the list.

If there are more than 2 labels the border is turned into a double border similar to before.

image

And three bookmark example....

image


tuomas2:
let's try! interesting...

PR #1402, opened 2021-07-24
Revert "Show two color underline"

tuomas2:
Reverts AndBible/and-bible#1397

Tried and this causes some bugs. Also quite hacky approach. Need to think about this a bit more still...


agrogers:
I actually quite like the look of the border highlight more than the normal underline. It underlines the numbers and cross references in a nice consistent way.

PR #1454, opened 2021-07-30
More application pref icons

agrogers:
Two more icons.


tuomas2:
There are now conflicts, can you resolve? Gave instructions via teams

PR #1456, opened 2021-07-31
Minor formatting changes

agrogers:
image

image

image

### DIFF
--- a/app/bibleview-js/src/common.scss
+++ b/app/bibleview-js/src/common.scss
@@ -160,3 +160,7 @@ $modal-content-background-color-night: black;
   width: 100%;
   overflow: hidden;
 }
+
+.small {
+    font-size: 0.9em
+}
--- a/app/bibleview-js/src/common.scss
+++ b/app/bibleview-js/src/common.scss
@@ -160,7 +160,3 @@ $modal-content-background-color-night: black;
   width: 100%;
   overflow: hidden;
 }
-
-.small {
-    font-size: 0.9em
-}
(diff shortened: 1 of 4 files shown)


app/bibleview-js/src/common.scss
```diff
@@ -160,3 +160,7 @@ $modal-content-background-color-night: black;
   width: 100%;
   overflow: hidden;
 }
+
+.small {
+    font-size: 0.9em
+}
```
tuomas2: Why not in AmbiguousSelectionBookmarkButton.vue?

tuomas2: better there as this definition is too generic ("small")

tuomas2: here should be only those definitions that are used more widely in many components

agrogers: That's fine. But I would use 'small' as a generic way of getting smaller text. I didn't find an existing class that we could use for that. Does one exist?

tuomas2: Why not putting it first in AmbiguousSelectionBookmarkButton and move it common when you find that you need that in multiple places? :-) Now it's only used in one place. But well, I'm not strict about this, if you want it to be in common, it's okay too.

tuomas2: I also kind of don't want unexpected effects to happen all around the app, if I happen to want to adjust one font size. What works in one context as "smaller" font, might not work very well in other? That's why I like Vue's scoped styling.

agrogers:
Sure. Makes sense. I will move it across now.

I like the em sizing option because it doesn't matter too much what the
context is - it just makes it smaller by the standard amount.  I probably
would have done journal-name a bit differently. It's size is 180%. I find
doing it directly like that makes it harder to keep sizes consistent. I
would have had a generic class like extra-large and added it to the
template.

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

PR #1527, opened 2021-08-07
Changed the style of bar buttons for extra clarity

agrogers:
The unpinned buttons looked the same as the pinned buttons (with the exception of tiny pin icon). 

The unpinned buttons function as a group so it seemed better to display them so that the look like a group and distinct from the other buttons. So i reduced the corners a little. I tried hard to not have to create new resource files for them but alas, I could not see how to manipulate just the corner style.

I also modified the New Window button to look less like the existing window buttons and more like the a button that might create something.

image


app/src/main/java/net/bible/android/view/util/widget/WindowButtonWidget.kt
```diff
@@ -84,6 +84,8 @@ class WindowButtonWidget(
     }
 
     private fun updateBackground() {
+        val isPinnedWindow = if (window?.isPinMode == null) false else window?.isPinMode
```
timbze:
```suggestion
        val isPinnedWindow = window?.isPinMode == true
```

agrogers: I had trouble with something like this because where I was using it require a Boolean and I think isPinMode is a nullable Boolean. I dont think i tried this though. I will try it.

agrogers:
Not sure what i was doing before but your change worked fine Tim. I fixed the line below it as well.

timbze:
> Not sure what i was doing before but your change worked fine Tim. I fixed the line below it as well.

Generally you don't need the "== true" part, but since you're also checking against null value in this case then it works like that.

agrogers:
Hi @tuomas2 . I quite like this visual change.  What do you think?

tuomas2:
Sorry this has slipped from my eyes. I'll check soon.

agrogers:
No problems. i am always impressed by how many bugs and features you can
add in one day!

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

tuomas2:
I'm not sure if I like, but I'll give it a try (with option to revert later) :-)

agrogers:
You will like it, I am sure of it :)

tuomas2:
It gave "unconsistent" feeling until I understood the logic. I think rounded corners to the edges (#1619) would help that

agrogers:
That feeling is exactly what i was hoping for so mission accomplished! I think rounded corners works against a little. Commented on your other Issue. My aim was to show that they are one group or act as one button so I think they should look different to the other buttons.

Same logic for the add button which i also modified - it is not a window button, it creates a window button - so i think it's appearance  should convey that.

PR #1541, opened 2021-08-07
Add maximised button to bookmark modal

agrogers:
I think this will be really handy. It allows the user to maximise the current window with a single click. Great when entering notes especially. But also useful when adding bookmarks if there are more than a few windows on screen.

Unfortunately it doesn't work properly. If you click the button twice the other windows go white. And if you then fullscreen it the maximised window is displayed correctly. I suspect it is not too hard to fix for someone who knows what he is doing. I am guessing that it must be the context from which i am calling the action - i am just calling the existing maximise function.

I am not real happy with the look of the icon. It is the correct thickness compared to other icons but not these ones. If you include this i will try to improve that.

Oh, one thing that I didn't do was a check to **not** show the maximise button when the window is maximised. Happy to try to do that if you are to include this pull request.

I also moved the **Extend Verse Selection** button to next to the verse reference. I had been thinking for a while that this might help convey that this button applied to the verses. The extra button kind of made that even more compelling.

image

image


agrogers:
How do i see what the lint and test errors are?

tuomas2:
open test results artifact zip file and read html or read test run log

tuomas2:
okay arrifact was not built so only log https://github.com/AndBible/and-bible/runs/3269882876?check_suite_focus=true

tuomas2:
I guess this also is intended to be "Ready for review"

tuomas2:
There are conflicts in this PR, they need to be resolved first. 

Also I'm afraid we are adding too many tool buttons to 1-click modal and this and current min/max button are quite close in essence what they do. It might be confusing to have two similar functions on toolbar with slightly different icons. Hmm.

tuomas2:
What about adding this feature only to bookmark modal? There are not too many buttons yet and that's where this feature would be most useful anyway.

agrogers:
Sure, good idea. It is the editing window that i really only use it on. I
like it.

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

agrogers:
I will look at this again sometime.

PR #1580, opened 2021-08-14
Reverse the date order of the bookmark list so most recent are at the top.

agrogers:
My list is now realllly long. It takes 5 seconds or more to scroll to the bottom of the bookmarks. But almost always I am looking for bookmarks I just created or created recently... which were at the bottom. Now they are at the top.

I didn't know how to add an extra sort option - Biblical Order, Date Order, Date Order Desc.

There is no string i could use for the latter anyhow. So i have just reversed the current one.

### DIFF
--- a/app/src/main/java/net/bible/android/view/activity/bookmark/Bookmarks.kt
+++ b/app/src/main/java/net/bible/android/view/activity/bookmark/Bookmarks.kt
@@ -268,7 +268,8 @@ class Bookmarks : ListActivityBase(), ActionModeActivity {
     private fun changeBookmarkSortOrder() {
         bookmarkSortOrder = when (bookmarkSortOrder) {
             BookmarkSortOrder.BIBLE_ORDER -> BookmarkSortOrder.CREATED_AT_DESC
-            BookmarkSortOrder.CREATED_AT_DESC -> BookmarkSortOrder.BIBLE_ORDER
+            BookmarkSortOrder.CREATED_AT_DESC -> BookmarkSortOrder.CREATED_AT            
+            BookmarkSortOrder.CREATED_AT -> BookmarkSortOrder.BIBLE_ORDER
             else -> BookmarkSortOrder.CREATED_AT_DESC
         }
     }
--- a/app/src/main/java/net/bible/android/view/activity/bookmark/Bookmarks.kt
+++ b/app/src/main/java/net/bible/android/view/activity/bookmark/Bookmarks.kt
@@ -60,7 +60,8 @@ val BookmarkSortOrder.description get() =
     when(this) {
         BookmarkSortOrder.BIBLE_ORDER  -> CommonUtils.getResourceString(R.string.sort_by_bible_book)
         BookmarkSortOrder.LAST_UPDATED -> CommonUtils.getResourceString(R.string.sort_by_date)
-        BookmarkSortOrder.CREATED_AT -> CommonUtils.getResourceString(R.string.sort_by_date)
+        BookmarkSortOrder.CREATED_AT -> CommonUtils.getResourceString(R.string.sort_by_date) + "🠕"
+        BookmarkSortOrder.CREATED_AT_DESC -> CommonUtils.getResourceString(R.string.sort_by_date) + "🠗"
         BookmarkSortOrder.ORDER_NUMBER -> "order number"
     }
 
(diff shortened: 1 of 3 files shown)


agrogers:
I timed how long it takes me to scroll to the bottom of my list today. 13 seconds! Would be great to add this or include an additional option on the sort button that does it.

tuomas2:
Is this ready to review?

agrogers:
Do you mean is it complete? If so, yes. I have only been adding pull requests that i think are ready.

tuomas2:
Okay, marking it 'ready for review' then :-)

db/src/main/java/net/bible/android/database/bookmarks/BookmarksDao.kt
```diff
@@ -35,6 +35,7 @@ import java.util.*
 const val orderBy = """
 CASE WHEN :orderBy = 'BIBLE_ORDER' THEN Bookmark.kjvOrdinalStart END,
 CASE WHEN :orderBy = 'BIBLE_ORDER' THEN Bookmark.startOffset END,
+CASE WHEN :orderBy = 'CREATED_AT_DESC' THEN Bookmark.createdAt END DESC,
```
tuomas2: I believe this should be

```suggestion
CASE WHEN :orderBy = 'CREATED_AT_DESC' THEN Bookmark.createdAt DESC END,
```

db/src/main/java/net/bible/android/database/bookmarks/BookmarksDao.kt
```diff
@@ -45,6 +46,7 @@ CASE WHEN :orderBy = 'BIBLE_ORDER' THEN Bookmark.kjvOrdinalStart END,
 CASE WHEN :orderBy = 'BIBLE_ORDER' THEN Bookmark.startOffset END,
 CASE
     WHEN :orderBy = 'CREATED_AT' THEN Bookmark.createdAt
+    WHEN :orderBy = 'CREATED_AT_DESC' THEN Bookmark.createdAt
```
tuomas2: And here I would feel that you meant

```suggestion
    WHEN :orderBy = 'CREATED_AT_DESC' THEN Bookmark.createdAt DESC
```

Or what do you think?

app/src/main/java/net/bible/android/view/activity/bookmark/Bookmarks.kt
```diff
@@ -266,9 +267,9 @@ class Bookmarks : ListActivityBase(), ActionModeActivity {
 
     private fun changeBookmarkSortOrder() {
         bookmarkSortOrder = when (bookmarkSortOrder) {
-            BookmarkSortOrder.BIBLE_ORDER -> BookmarkSortOrder.CREATED_AT
-            BookmarkSortOrder.CREATED_AT -> BookmarkSortOrder.BIBLE_ORDER
-            else -> BookmarkSortOrder.CREATED_AT
+            BookmarkSortOrder.BIBLE_ORDER -> BookmarkSortOrder.CREATED_AT_DESC
+            BookmarkSortOrder.CREATED_AT_DESC -> BookmarkSortOrder.BIBLE_ORDER
+            else -> BookmarkSortOrder.CREATED_AT_DESC
```
tuomas2: Shouldn't this be full loop between BIBLE_ORDER -> CREATED_AT_DESC -> CREATED_AT -> BIBLE_ORDER? Now we don't go ever to CRATED_AT.

tuomas2:
```suggestion
            BookmarkSortOrder.BIBLE_ORDER -> BookmarkSortOrder.CREATED_AT_DESC
            BookmarkSortOrder.CREATED_AT_DESC -> BookmarkSortOrder.CREATED_AT            
            BookmarkSortOrder.CREATED_AT -> BookmarkSortOrder.BIBLE_ORDER
            else -> BookmarkSortOrder.CREATED_AT_DESC
```

tuomas2 (changes requested):
Check my suggestions and apply them if you think they are correct. Please also test quickly if they work as intended.

agrogers:
Um, i know my original pull request works - i have been using it on.

And I remember getting confused by where the END of the case should be
placed - i reckon i have it right.

However you are correct in that my pull request never sorts by Created_At
in an ascending order. I mentioned that in the original comment I think. I
didn't know how to add that as an extra option in the button press loop.
But it seems like you have done it and it was much easier than i realised.

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

tuomas2:
@agrogers here's the syntax: https://sqlite.org/lang_expr.html . It can indeed work either way but have side-effect if placed wrongly.

agrogers:
Thanks for the link. Happy to look at this again but cant do it until the
weekend.

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

tuomas2:
Okay I made changes and will merge and check if it works...

agrogers:
Sweet. Thank you

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

PR #1660, opened 2021-09-01
Created separate logo for 'debug' build type

uniqueid001:
Created a separate logo to easily distinguish between the official
released build and a debug build that is also installed.
Additionally, append the text "*Debug*" to the app's name.


tuomas2:
I believe it might be possible to do this without copy-pasting whole drawable file. I'll see that option before merging.

tuomas2:
Changed my mind. We are not changing logo that often... so let it be this way.

uniqueid001:
: )

That was my perspective too.  I started using a Layer Drawable to put the two together.  That would require renaming the original drawable, creating an alias in main, creating the overlay drawable in debug, and then creating the layer drawable in debug.  

It just so happened that I actually did this a week prior but hadn't submitted it and the next day you implemented a change to the logo.  So I had to update it with the new paths.  It turned out to be quite quick and simple to do.

PR #1734, opened 2021-09-21
Check language files

### LINKED ISSUE #1003: Automatic checking for js side translations
Need to check that translation files have %s etc in correct places as transifex does not check them. Otherwise we will have js side errors as a result... 

Python script that is executed when doing make tx-pull will do fine.

JJK96:
Adds a python script to `make tx-pull` to check the language files for the correct number of placeholders.
Fixes #1003 

The script gives an output like the following if a file is incorrect:
```
$ python check.py
Traceback (most recent call last):
  File "/home/jjk/git/and-bible/app/bibleview-js/src/lang/check.py", line 49, in <module>
    check_counts(counts, default_counts)
  File "/home/jjk/git/and-bible/app/bibleview-js/src/lang/check.py", line 41, in check_counts
    raise CountsNotMatch(string, count, target_count)
__main__.CountsNotMatch: Error in /home/jjk/git/and-bible/app/bibleview-js/src/lang/ru.yaml: refParserHelp contains 0 placeholders, while it should contain 1 placeholders
```


tuomas2:
Thank you for implementing this! 

I fixed ru and pt translations in transifex too.

PR #1768, opened 2021-09-28
Improve 3.3 error report

timbze:
Add webview and sqlite versions to error report in 3.3

### DIFF
--- a/app/bibleview-js/.gitignore
+++ b/app/bibleview-js/.gitignore
@@ -1,0 +1,23 @@
+.DS_Store
+node_modules
+dist
+
+
+# local env files
+.env.local
+.env.*.local
+
+# Log files
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+pnpm-debug.log*
+
+# Editor directories and files
+.idea
+.vscode
+*.suo
+*.ntvs*
+*.njsproj
+*.sln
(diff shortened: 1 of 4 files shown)


tuomas2:
What motivates this?

timbze:
In case 3.3 would get another update, it would be easier to get which webview version the user has.

The .gitignore is because when I switch to 3.3 branch it always shows lots of files that are ignored in 4.0

I was thinking of also doing a PR for automatic db restore (the backed up version from 3.3) if user wants to downgrade, but not sure that I'll have time for that.

timbze:
But according to test it actually looks like this is not working, so I'll have to check a bit more first.

tuomas2:
How would it work? What would be benefit over the current downgrade path?

timbze:
> How would it work? What would be benefit over the current downgrade path?

That would make it possible to install 3.3 over 4.0 and automatically get downgraded. Instead of doing migration, app would check db before Room accesses it, and automatically restore the 3.3 backup file (before starting app essentially).

tuomas2:
Android prevents downgrades. 4.0 needs always to be uninstalled first.

timbze:
> Android prevents downgrades. 4.0 needs always to be uninstalled first.

Oh you're right. Good to know

PR #1792, opened 2021-10-03
Moved the 'saved' indicator to the top.

agrogers:
I find the '-saved-' indicator is often over the top of the text i am typing because it is at the same height as the last line of text. I have moved it to the first line of text which means it will rarely overlap what i have recently typed.


tuomas2:
Let's try this.

PR #1793, opened 2021-10-03
Decreased the length of the line separating the note from the verse

agrogers:
I find it a little difficult to tell the difference between a note that is attached to a verse and a note that is attached to the study pad. The shorter line makes this distinction clear.

image

### DIFF
--- a/app/bibleview-js/src/common.scss
+++ b/app/bibleview-js/src/common.scss
@@ -128,10 +128,9 @@ $button-grey: #8e8e8e;
 
 .separator {
   border-width: 1pt 0 0 0;
-  border-style: dashed;
-
+  border-style: dotted;
   height: 1pt;
-  width: calc(100% - 10pt);
+  width: calc(33% - 10pt);
   margin: 10pt 5pt 5pt;
   border-color: rgba(0, 0, 0, 0.15);
   .night & {


tuomas2:
Hmm, it's not super pretty to my eyes. Could there be another option?

agrogers:
I recall you saying that about the padding once 😉

It reminded me of the way Word did it's end notes or footnotes or
something.

Maybe centre the smaller line?

tuomas2:
Try centering it and post screenshot :-) And maybe leave it a bit longer, like 60%...

agrogers:
Sure thing. Might take a week or two. In the meantime here is a screenshot of small company who has done something similar ;-)
image

tuomas2:
Hmm. That's ok for footnotes in printed/printable document. But here in study pad it is attached to all bookmarks, so it looks weird to me.

agrogers:
You could be right - I didn't spend much time on it actually :)

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

agrogers:
Centre looks ok for longer notes (which admittedly most probably are).  But the line doesn't feel so connected to the note. If the line get's too long then it feels like the note is not attached to the verse which was my reason for visiting this in the first place. But a shorter line either left aligned or center aligned is good with me. Maybe center is best :)

image
image

tuomas2:
Still I don't like it. Let me try to think about another solution... :-)

agrogers:
Sure thing.

tse-gratis:
italic?

tuomas2:
I'm trying now to make it simply much thinner. Let's try.

PR #1805, opened 2021-10-05
migrate gradle build scripts from groovy to kotlin

uniqueid001:
.


tuomas2:
- [ ] Related: https://github.com/AndBible/jsword/pull/6

app/build.gradle.kts
```diff
@@ -0,0 +1,307 @@
...
+    }
+    return stdout.toString().trim()
+        }) // use()
+}
+
+val npmUpgrade by tasks.registering(Exec::class) {
+    inputs.file("$jsDir/package.json")
+    outputs.file("$jsDir/node_modules/.bin/npm")
+    workingDir = file(jsDir)
+    // Workaround for F-droid, which has buggy npm version 5.8, that always fails when installing packages.
```
tuomas2, line 44: Node is updated to version 14.16 in F-droid, and there should be more recent npm
 https://gitlab.com/fdroid/fdroiddata/-/blob/master/metadata/net.bible.android.activity.yml

tuomas2: ah, just noticed that this workaround is also in old version...

tuomas2:
@uniqueid001 check the improvements made in the above mentioned commit ^

tuomas2:
and thank you for doing this! This change has been in my mind. Good that we got it done now. I understand kotlin much better than groovy so it is easier to maintain.

app/build.gradle.kts
uniqueid001: : )
Yeah, I simply translated it.

uniqueid001:
> @uniqueid001 check the improvements made in the above mentioned commit ^

I missed the nullable in the cast;  it compiled and I have the property defined so it was non-null at runtime too.
I see a number of style changes.  I am just learning kotlin myself, as I teach it to a homeschool class.  I am not accustomed to things like using an 'if' as an expression instead of a statement, or assigning a single expression to a function declaration.

Issue #1070, opened 2021-05-16
Enable Swiping up on a window button to maximise the window 


agrogers:
I am using the maximise window option a lot more now. It feels natural to just click and swipe up on the window button I want to maximise.

(the action I am trying to describe is the same action to bring down the workspace selector - click and swipe down in that case).

agrogers:
Easy to implement i think but...

I suggest we delay this one to after v3.4 release

agrogers:
Completed with https://github.com/AndBible/and-bible/pull/1888
I use this all the time now. I like it.

tuomas2:
I also like it and have been using it.

PR #1834, opened 2021-10-11
Enabled swipe up and down on window buttons for maximise and minimise

agrogers:
This addresses this request: https://github.com/AndBible/and-bible/issues/1070

I often want to minimise or maximise a window and I find both longpress and accessing the popup menu a little intrusive in my work flow. 

This change enables that functionality using swiping down and up which I have found really nice. I copied Tuomas' code from the workspace fling controls. It feels like my code should be in some more generic location that could be reused by other areas. But that is beyond me.

The only thing I don't like is that it is not immediately clear what has happened. Window animation would resolve that but that sounds tricky. Most people will not find this functionality unless it is promoted so it wont affect people.


agrogers:
Hmm, it seems I broke the popup window when clicking the window button. I will have a look at that on the weekend and try again.

PR #1835, opened 2021-10-11
Grouped book buttons when displayed horizontally

agrogers:
I have been using this app for a long time but I have still found it difficult sometimes to locate the book that I want. I think the reason is that I need a clearer grouping system. The colors are supposed to do that. But I have found that insufficient. There are some obvious groupings - for example, I think Matthew should start on a new line not in the middle of the row.

I have just commandeered the existing horizontal option which means the old option doesn't exist anymore. Perhaps this display format should be provided as an additional option. But I kind of see it as an improved version.

This display option is obviously incompatible with the alpha-ordered option and vertical arranged option.

image


timbze:
1. I like this
2. I think the alpha-order option should definitely remain
3. I'm not totally convinced that we should just switch it to this, maybe try in beta when going to 4.1 and see how less users respond first

timbze:
One other question on my mind, how about landscape? Landscape only shows 6 rows currently... one option would be to put in a ScrollView so that you would have to scroll down to get to Revelation

agrogers:
My landscape mode worked but the buttons were very thin. Not sure how to
remove the padding.

tuomas2:
Looks good and might work well.

- [ ] Shouldn't there be a switch (in 3-dot menu) for enabling / disabling this?

agrogers:
> Shouldn't there be a switch (in 3-dot menu) for enabling / disabling this?

Probably but I didn't know how to do that :)

tuomas2:
You could see example how CommonUtils.settings is used in GridChoosePassageVerse.kt for the (rather) new sorting feature

app/src/main/java/net/bible/android/view/activity/navigation/GridChoosePassageBook.kt
```diff
@@ -16,6 +16,7 @@
  *
  */
 
+// TODO: Not saving the menu option. Don't know how to set the other menu options when one option is changed.
```
tuomas2: - [ ] TODO

tuomas2: It looks like it is saved so this should be simply removed

```suggestion
```

app/src/main/java/net/bible/android/view/activity/navigation/GridChoosePassageBook.kt
```diff
@@ -65,6 +66,9 @@ class GridChoosePassageBook : CustomTitlebarActivityBase(R.menu.choose_passage_b
     @Inject lateinit var navigationControl: NavigationControl
     @Inject lateinit var activeWindowPageManagerProvider: ActiveWindowPageManagerProvider
 
+    data class ExtraBookInfo(val Color: Int, val GroupA: String, val GroupB: String)
+    private lateinit var mmenu: Menu
```
tuomas2: mmenu or menu?

app/src/main/java/net/bible/android/view/activity/navigation/GridChoosePassageBook.kt
tuomas2: You just save as normal, and then invalidate options (see my commits for how to do it). I think it is done correctly now, right?

tuomas2:
I guess I'll merge this too. Looks ok and works.

agrogers:
Hmm, I have been using it for a while now on my own build so yes, I think
it is working. I suspect I learnt how to solve my problem but forgot to
remove the TODO. I will check your changes.

PR #1867, opened 2021-10-18
Moved the 'saved' notice behind the text and changed it to an icon

agrogers:
Changed the z-index to get the message behind the text and swapped the 'Saved' message with a saved icon.

image

image

### DIFF
--- a/app/bibleview-js/src/components/TextEditor.vue
+++ b/app/bibleview-js/src/components/TextEditor.vue
@@ -36,7 +36,7 @@
   </InputText>
   <div @click.stop class="edit-area pell">
     <div ref="editorElement"/>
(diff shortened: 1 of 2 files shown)


tuomas2:
Getting better! Not sure if that old discette icon is best for this. It does not seem obvious to me (especially that small).

tuomas2:
And I'd try grey or blue color.

agrogers:
Yeah, only the old guys know about discs. But it seems to still be the
standard for 'save'.

I was actually thinking of removing it all together. It is really not
needed. It just gives some peace of mind and i guess tells people they dont
have to push a save button.

I didn't go for gray because that  conveys the idea that it is a button to
be pressed. Green usually means all good. You can change it to whatever you
like though. Invisible is fine now too :)

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

PR #1883, opened 2021-10-23
Simplify Github release CI code

timbze:
### Reason for this
- `actions/create-release` is no longer maintained
- Simplify the yml code

### Note
I have not tested it here, but used the same way I did in other personal code

### DIFF
--- a/.github/workflows/build-apk.yml
+++ b/.github/workflows/build-apk.yml
@@ -81,32 +81,18 @@ jobs:
         echo ${{ secrets.KEYSTORE_BASE64 }} | base64 --decode > keystore.jks
         $ANDROID_HOME/build-tools/29.0.3/apksigner sign --ks keystore.jks --ks-pass 'pass:${{ secrets.KEYSTORE_PASSPHRASE }}' --key-pass 'pass:${{ secrets.KEY_PASSPHRASE }}' --out $GITHUB_WORKSPACE/andbible-$TAG_NAME.apk app-github-release-unsigned.apk
         rm -f keystore.jks
+
     - name: Create release
       if: contains(github.ref, 'refs/tags/')
-      id: create_release
-      uses: actions/create-release@v1
+      uses: softprops/action-gh-release@v1
       env:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-apk.yml
```diff
@@ -81,33 +81,19 @@ jobs:
         echo ${{ secrets.KEYSTORE_BASE64 }} | base64 --decode > keystore.jks
         $ANDROID_HOME/build-tools/29.0.3/apksigner sign --ks keystore.jks --ks-pass 'pass:${{ secrets.KEYSTORE_PASSPHRASE }}' --key-pass 'pass:${{ secrets.KEY_PASSPHRASE }}' --out $GITHUB_WORKSPACE/andbible-$TAG_NAME.apk app-github-release-unsigned.apk
         rm -f keystore.jks
+
     - name: Create release
       if: contains(github.ref, 'refs/tags/')
-      id: create_release
-      uses: actions/create-release@v1
+      uses: softprops/action-gh-release@v1
```
tuomas2, line 87:
```suggestion
      uses: softprops/action-gh-release@6034af24fba4e5a8e975aaa6056554efe4c794d0
```

tuomas2: How can we trust this author? I feel a bit scared always using these third-party modules that have minified js code... almost would like to build code myself from source and review code.

timbze: Good question. It was one of the 4 actions recommended from [Github's action repo](https://github.com/actions/create-release#github-action---releases-api). It was the one with most stars and forks, and looks like it's being constantly maintained.

tuomas2: It's pity that Github does not take responsibility of maintaining this kind of useful basic tools...

timbze: I'm just seeing that the source files are TypeScript. Isn't the minified js file just TypeScript processed? We could quite easily fork and do that process ourselves I think

tuomas2:
I'm not sure if this should be closed even if we can't merge it directly because I'm afraid old code might stop working some time in the future (contains deprecating api).

PR #1884, opened 2021-10-23
Study pad export 

### LINKED ISSUE #1328: export studypad in html, txt, even pdf if possible

tuomas2:
Closes #1328

### DIFF
--- a/app/bibleview-js/package-lock.json
+++ b/app/bibleview-js/package-lock.json
@@ -36,6 +36,7 @@
         "jsdom": "~16.6.0",
         "json-loader": "^0.5.7",
         "npm": "~8.0.0",
+        "raw-loader": "^4.0.2",
         "sass": "~1.29.0",
         "sass-loader": "^10.2.0",
         "vue-cli-plugin-webpack-bundle-analyzer": "~4.0.0",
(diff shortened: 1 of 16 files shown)


agrogers:
Look forward to seeing it.

tuomas2:
ready on Thursday probably

tuomas2:
Waiting for some help https://stackoverflow.com/questions/69750225/how-to-import-raw-css-as-raw-text-in-js-side-generated-from-vue-components

tuomas2:
Screenshot_20211028_170344

tuomas2:
Exporting kind of works but it does not have CSS yet, so it looks somewhat ugly.

tuomas2:
feedback about this @AndBible/developers ?

timbze:
I'm happy with the format, but would be nice to be able to copy as html (to paste in email etc), not sure if that's possible.

tuomas2:
I agree. But already now it is easy to upload to Google Docs and Google Docs can be used even to edit the document.

JJK96:
Looks fine to me, indeed a copy to clipboard would be nice-to-have, but this is a nice addition already!

PR #1887, opened 2021-10-25
Show the 'to' verse for the selected range

agrogers:
When multiple verses are selected the range is shown (eg Hebrews 6:13-15) not just the first verse (Hebrews 6:13)

It doesn't handle cross chapter or book boundaries properly - just increments the to verse number. But I feel like this is still better than nothing.

image

### DIFF
--- a/app/bibleview-js/src/components/modals/AmbiguousSelection.vue
+++ b/app/bibleview-js/src/components/modals/AmbiguousSelection.vue
@@ -55,7 +55,7 @@
     </div>
     <template #title>
       <template v-if="verseInfo">
-        {{ bibleBookName }} {{ verseInfo.chapter}}:{{verseInfo.verse}}
+        {{ bibleBookName }} {{ verseInfo.chapter}}:{{verseInfo.verse}}<template v-if="verseInfo.verseTo">-{{verseInfo.verseTo}}</template>
       </template>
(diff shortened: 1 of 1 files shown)


timbze:
Then I'm guessing it will also be an issue when there's a missing verse (which some versions have).

I think either that should be handled properly, including cross chapter,

Or, show something like "Hebrews 6:13 +2"

agrogers:
Yep, I agree. I was going to do the +2 approach but that was less helpful to me and I think still fails because there is a verse 0 hidden at the top of the chapter. I don't know how to handle the cross chapter scenarios so I will have to leave that to others or get some suggestions. It works well enough for me on my build so I am happy :)

tuomas2:
Perhaps we can add just "+" so it does not lie about amount of verses but indicates that "there's more".

agrogers:
If it is a lengthy passage like sometimes happens at church I want to know
that I have selected the right range. But I can't always see the last
verse. Showing + is better than nothing but I prefer to see the verse it is
going to. I very rarely select past a chapter end so it is not really an
issue for me.

tuomas2:
I mean + *if* it exceeds chapter boundary. For example: John 3:16-36+

tuomas2:
@agrogers check 961b285

PR #1888, opened 2021-10-25
Swipe action on window buttons to maximise and minimise windows

agrogers:
Window buttons (top right of the window) can now be swiped up and down to maximise and minimise the window. 

I use it quite often to quickly maximise a window rather than having to go through the popup menu. It feels much more natural now.

I use it less for minimising because the button bar at the bottom can do that as well. But it is still handy.

### DIFF
--- a/app/src/main/java/net/bible/android/view/activity/page/screen/BibleFrame.kt
+++ b/app/src/main/java/net/bible/android/view/activity/page/screen/BibleFrame.kt
@@ -21,10 +21,14 @@ package net.bible.android.view.activity.page.screen
 import android.annotation.SuppressLint
 import android.os.Build
 import android.util.Log
+import android.view.GestureDetector
 import android.view.Gravity
+import android.view.MotionEvent
 import android.view.View
+import android.view.ViewConfiguration
 import android.view.ViewGroup
 import android.widget.FrameLayout
+import androidx.core.view.GestureDetectorCompat
 import net.bible.android.BibleApplication
 import net.bible.android.control.event.ABEventBus
 import net.bible.android.control.page.window.Window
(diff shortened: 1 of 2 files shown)


app/src/main/java/net/bible/android/view/activity/page/screen/BibleFrame.kt
```diff
@@ -132,31 +138,74 @@ class BibleFrame(
...
-            window
-        )
-    }
+//    private fun createCloseButton(window: Window): WindowButtonWidget {
+//        val text = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) "☰" else "="
+//        return createTextButton(text,
+//            { v, motionEvent -> windowControl.closeWindow(window); true},
+//            window
+//        )
+//    }
```
tuomas2: Commented out code should not be left if there's no specific reason for it

tuomas2:
```suggestion
```

app/src/main/java/net/bible/android/view/activity/page/screen/BibleFrame.kt
```diff
@@ -112,7 +118,7 @@ class BibleFrame(
         val button =
             when {
                 isSingleWindow -> return
-                window.isLinksWindow -> createCloseButton(window)
+//                window.isLinksWindow -> createCloseButton(window)
```
tuomas2:
```suggestion
```

tuomas2:
First impression is good. Let's try...

agrogers:
Yeah, I guess it is easy enough to see what code is removed.

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

tuomas2:
@agrogers, see 3267d47

agrogers:
I use the swipe down quite often too but i find that i often start my swipe on the dividing line so i reduce the size of the window rather than minimising. I was going to try shifting the button up a few pixels to see if it solved that problem (although i think it might look a little ugly). Any thoughts on how to avoid that?

tuomas2:
I think the area where line takes input should be reduced to exclude window button proximity. I don't know exactly what is best way to do that.

Issue #611, opened 2020-04-12
Change the window the search results appear in


agrogers:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

tuomas2:
I like this idea. Need to think about it further though because new document should be stored somehow (so that it does not disappear when quiting/ switching apps).

agrogers:
Re killing the app, the existing search window dies with the app if it is left open so the new approach would be no worse.

I feel like an approach might be to always have a search 'bible' in the Bible drop down selector. The results from the last search get loaded into here. So you could always go back to the results of the last search if that list is stored somehow.

If the search 'bible' is open then the user sees the search results loaded after he closes the standard search window. If not then everything just works exactly as it does. That's good for existing users since nothing changes and you won't get complaints. 

Assuming the Search bible is open, we now have a bible with a subset of verses which functions like any other Bible. Not sure about syncing though. It is easy enough if other stuff is synching to it. But what about it syncing to other things. I originally imagined it like the commentaries which often don't have anything to show for a particular verse. But they just show one verse at a time. The Bible search window is showing all search result verses. Maybe it can just work the same way that the New Testament only works when showing the Old Testament verse - 'Not found in document'.

But if it did how would you get your search results to show again? Perhaps all verses get redisplayed in the search window when it receives focus. 

Another question is what happens when the user clicks the verse selector button while the search window is active. Initially I liked the idea of jumping direct to the new search dialog to modify or do a new search.

But you already have the neat feature that only shows the books of the Bible available for that bible. That would be perfect for large searches. I don't have to scroll through 100 verses to get to Malachi. Does the functionality also extend to chapters or is it assumed that if the book is there then all chapters and all verses will also be there? If it did then it would be super easy to jump directly to the exact verse in your list without scrolling at all. 

Hmm, the buttons get very big when there are not many books or chapters to show. That means there is plenty of space in the button to show the number of search hits in each book and maybe even each chapter. I like it. 

Anyhow, I'm sure there are a pile of other restrictions and issues that need to be thought about here. But it does seem like And Bible already has almost the perfect structure to take the search from rudimentary to awesome in one biggish step 🙂

timbze:
Additional idea from #899 - add feature to copy verse from search results list

tuomas2:
Now would be easy to implement this, showing as MultiDocument.

agrogers:
[...]

PR #1938, opened 2021-11-08
Highlight the words in the search results

agrogers:
It handles multiple words and the '*' operator. It doesn't handle phrases or '?' yet. And it is not tested. Let me know if it is heading in the right direction and I (or you) can finish it off later.

image

image

### DIFF
--- a/app/src/main/java/net/bible/android/control/search/SearchControl.kt
+++ b/app/src/main/java/net/bible/android/control/search/SearchControl.kt
@@ -56,6 +56,7 @@ class SearchControl @Inject constructor(
     )
 {
     private val isSearchShowingScripture = true
+//    public final var originalSearchString = ""
 
     enum class SearchBibleSection {
         OT, NT, CURRENT_BOOK, ALL
@@ -106,6 +107,7 @@ class SearchControl @Inject constructor(
 
         // add search type (all/any/phrase) to search string
         decorated = searchType.decorate(cleanSearchString)
+        originalSearchString = decorated
 
         // add bible section limitation to search text
         decorated = getBibleSectionTerm(bibleSection, currentBookName) + " " + decorated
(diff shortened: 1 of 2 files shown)


timbze:
I'm guessing it won't handle strongs search either? (haven't looked at code yet)

For example, would be very handy when clicking "find all references" on a Greek/Hebrew dictionary definition if the result words would highlight

agrogers:
Nope. But it is on my list to look at. That is probably the feature i most
want in this list.

These changes are very primitive. It should be replaced with the results
that come back from the Lucene queries but that needs some deep work
because it doesnt send back any info on the matches just the verse
references. But i think what I have there now would cover 95% of what most
people search for. Maybe more.

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

tuomas2:
Just a quick comment for now: I would prefer this is implemented first: https://github.com/AndBible/and-bible/issues/611 because it changes how this needs to be implemented.

agrogers:
611 is beyond my expertise at the moment. But I think there would still be
a place for the current simple search. Having become more familiar with our
users, I can see plenty of tickets coming through with 'what happened to
the old search - it was simple and this new one is too complicated'.

My thought had been that the simple search results list should remain with
the inclusion of a button at the top to open those results as document
window.

I think I would use both. Hard to know though until I have the new search
display in my hands.

timbze:
Depends how soon or if someone is planning to work on #611, it might be a good idea to implement this for now. I agree that it's better than nothing, plus, it's already done. All that if the code is ok. I have not reviewed the code.

tuomas2:
#611 is straightforward to implement, now that we have the "framework" in place. Not easy, requires knowing the current system of course, but nothing "new" there. It would be beneficial if someone else would do it as it would spread the knowledge. I am available for help. 

> 611 is beyond my expertise at the moment. But I think there would still be
> a place for the current simple search. Having become more familiar with our
> users, I can see plenty of tickets coming through with 'what happened to
> the old search - it was simple and this new one is too complicated'.

If I had that mindset I think I had to stop coding. I guess that's what certain users wish for sure. 
But that's not right attitude I think. Changes (that cause 1 star ratings from certain users) are 
necessary if we want to improve things.

That being said, if nobody is interested in taking the challenge of #611, this "old approach" can be 
acceptable too.

agrogers:
This method has been superseded by https://github.com/AndBible/and-bible/pull/1953 
This one is a lot simpler and maybe faster since it doesn't have to process each word before applying the highlighting. Anyhow, I will close it in favour of the other.

PR #1942, opened 2021-11-11
Increase default left/right margin to 3 (was 0)

uniqueid001:
This change should provide some usability benefits, by default:
 * provides a visual aesthetic while reading
 * always makes an empty region available to enter/leave full-screen
    mode without opening the quick-action menu
 * will keep the text out of the curved area of screens which extend to
    the edge of the device


tuomas2:
Probably good idea! I always keep some margins, even larger than 3.

Issue #186, opened 2018-12-30
Improved search feature


tuomas2:
From user:

> When using the search function, is it possible to limit your results to the gospels, history books for example? Also what about defining your own books for a section (eg choosing books I want to be included in a search)?

steelwil:
Use Soundex

Example: if I search for colour it should return results for colour and color

tuomas2:
From #202 

> Make the find input text box to contain the last 10 prev word searched. a down arrow icon to the right if the box to pop up the prev word/text searches.

huperniketes:
Also, add fields (eg, Strong's number for the word(s) in the search term(s)) and columns to hold them, which can then be prioritized for sorting.

josephjq:
thanks for working on this tuomas2.  i am the guy who emailed you that suggestion which you graciously added to the list.  i created this github account just for this so please forgive my newbishness.

i have continued to regularly use the app to learn more greek and hebrew, and i think i have a few more useful suggestions which might make sense to include with this feature, or a later one.  i have added them in brackets and italics to the feature list below because some are related to certain items on the list already.  hope this helps, and please correct me if i should be following a different format for this kind of community.




> From user:
> 
> > When using the search function, is it possible to limit your results to the gospels, history books for example? Also what about defining your own books for a section (eg choosing books I want to be included in a search)? _(it would also be very helpful to narrow strong's searches by including the specific morphology codes associated with a strong's number in a any one specific occurrence, and or by including more than one greek or hebrew Word in the search bar as a Phrase.)_
> 
> Other ideas:
> 
> * highlight search works in search results
> * Highlight the word related to strongs search in search results (see [1])  _(this would be helpful if the grammatical morphology codes could appear beside the highlighted Word when available and linked within the search list)_
> * help button in search dialog
> * quick search from toolbar
> * after large query, there should be 'wait' indicator before results are available _(better than a wait indicator would be loading the first 10 results immediately and then the rest in the background while the  reader begins with the first 10 or so.)_
> * Better search results for commentaries _(now they show only beginning of commentary text)_
> * search history _(yes please)_
> 
> From user [1]:
> 
> > i would like to request a feature which would make your app vastly more useful for me, and hopefully for others too. when i look at all the occurrences of a particular greek or hebrew Word from strong's, it would be extremely helpful to have the Word or Phrase which is translated from the particular greek or hebrew Word that i am looking up highlighted so i can be certain which Word of Phrase was used in each case of translation.

[...]

PR #1946, opened 2021-11-13
Show number of search results in window title

timbze:
Needs 1 new string

Relates to #186

But might be in same state as #1938 that might be useless after #611 is implemented

The "(2 results)" in title is the change in this PR
Screenshot from 2021-11-13 11-09-43


agrogers:
I like it. What happens if it hits the 1000 verse limit? Just shows 1000?

timbze:
> I like it. What happens if it hits the 1000 verse limit? Just shows 1000?

Yes it would just show 1000

PR #1953, opened 2021-11-22
Search highlighting including STRONGS 'Find all occurrences' 

agrogers:
This now handles all the highlighting options I think. Phrases, wild card chars and strong numbers. It seems the strong numbers differ in their formatting so it was necessary to manipulate them a little.

This method gets back the raw XML from the search instead of the plain text. This is unnecessary for non-Strongs searches but I think the overhead of converting to plain text has just been moved into the search routines.

It would also be simple to incorporate morphology searches now too.

I haven't tested it much. I will do so this week and see how it goes. I am sure the code can be much cleaner (and harder to understand ;-). Let me know if you want to try and do that and give me some hints where to start.

image


tuomas2:
Would you write some tests for those functions? They look scary :-)

agrogers:
Hi @tuomas2. I merged develop and this test fails now. The test you ignored fixed my other PRs though. 
```
net.bible.android.control.page.window.WindowControlTest > testGetMinimisedWindows FAILED
    kotlin.io.FileAlreadyExistsException: /tmp/robolectric-Method_testGetMinimisedWindows10211884927590351990/net.bible.android.activity-dataDir/databases/andBibleDatabase.db -> /tmp/robolectric-Method_testGetMinimisedWindows10211884927590351990/net.bible.android.activity-dataDir/files/database_backups/dbBackup-0-20211231-073632.db: The destination file already exists.
```

tuomas2:
> Hi @tuomas2. I merged develop and this test fails now. The test you ignored fixed my other PRs though.
> 
> ```
> net.bible.android.control.page.window.WindowControlTest > testGetMinimisedWindows FAILED
>     kotlin.io.FileAlreadyExistsException: /tmp/robolectric-Method_testGetMinimisedWindows10211884927590351990/net.bible.android.activity-dataDir/databases/andBibleDatabase.db -> /tmp/robolectric-Method_testGetMinimisedWindows10211884927590351990/net.bible.android.activity-dataDir/files/database_backups/dbBackup-0-20211231-073632.db: The destination file already exists.
> ```

I've noticed this kind of failure has been happening sporadicly every now and then lately. Not always. I have restarted the build always with success. But should take a look what is the real issue.

app/src/main/java/net/bible/android/control/search/SearchControl.kt
```diff
@@ -229,6 +250,7 @@ class SearchControl @Inject constructor(
         get() = isSearchShowingScripture || !currentDocumentContainsNonScripture()
 
     companion object {
+        lateinit var originalSearchString: String
```
tuomas2, line 227: This feels a bit unstable. It would be better to do this without this, feels hack.

tuomas2:
Did some code improving and cleaning up in c1ea356. 

Merging to develop.

agrogers:
Great. I will check it out. But...

I had to move all this code out of the adapter so I could access it while preparing the results to build book and word statistics. I think it is better outside the adapter - that didn't seem to be the place for it.

But I will take your code changes into my new attempt.

Also, I think people will really appreciate have highlighting there so good to merge it now.

PR #1980, opened 2021-12-13
Gave status bar a colour and made full screen mode immersive

### LINKED ISSUE #1979: Fix inconsistency in the status bar theme and make fullscreen mode immersive
# Issues
1. I find it inconsistent and irking that the status bar doesn't assume the same colour as the top bar of the application. In this screenshot, for example, the top bar is grey and the status bar is black. Every modern Android app, from Telegram, AOSP apps, Signal, to Youversion Bible theme the status bar accordingly.
 
Screenshot_20211212-201932_Bible Study

2. Non immersive full screen mode. Removing the black bar that serves no apparent purpose in full screen mode would be a great step towards improving the UX of the app.

Screenshot_20211212-212053_Bible Study

[StackOverflow question that has a solution to the full screen bar problem (idk if it applies to AndBible)](https://stackoverflow.com/questions/56353716/how-to-remove-top-status-bar-black-background)

Lastly, It would be great if AndBible could pull colours from the Android 12 Wallpaper theme. I understand if this is undesirable for a need to maintain brand identity or something, or because it would affect backwards compatibility.

I would love to work on these issues and make a PR, but unfortunately, I have no experience with Android development, Java or Kotlin.

actuday6418:
Completely new to Android development. I have no idea if this has any unintended consequences. Fixes #1979

### Status bar is themed grey like the top bar
Screenshot_20211213-144019_Bible Study _Debug_

### Full screen is immersive, no black bar on top
Screenshot_20211213-150655_Bible Study _Debug_

### DIFF
--- a/app/src/main/res/values-night/colors.xml
+++ b/app/src/main/res/values-night/colors.xml
@@ -28,6 +28,8 @@
     <color name="bar_window_button_background_colour_visible">#6A6A6A</color>
     <color name="bar_window_button_stroke_color">#686868</color>
 
+    <color name="status_bar">@color/black</color>
+
     <color name="window_button_text_colour">#32FFFFFF</color>
 
     <color name="navigation_bar">@color/black</color>
--- a/app/src/main/res/values-v27/barstyles.xml
+++ b/app/src/main/res/values-v27/barstyles.xml
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<resources xmlns:tools="http://schemas.android.com/tools" xmlns:android="http://schemas.android.com/apk/res/android">
+    <style name="BaseAppTheme" parent="@style/Theme.AppCompat.DayNight.DarkActionBar">
+        <item name="actionMenuTextAppearance">@style/MyMenuTextAppearance</item>
+        <item name="actionButtonStyle">@style/MyActionButtonStyle</item>
+        <!-- Prevent sub menu items (e.g. Admin sub-menu) appearing as 'white on white' on Android 2.3 -->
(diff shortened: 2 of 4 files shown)


tuomas2:
check night mode too

actuday6418:
Okay

actuday6418:
The status bar now becomes black in night mode. There is a problem, however: the bar stays black when night mode is switched off. An app restart is required to get it into the right colour (grey). Turning on night mode works correctly without restarting the app though.

I found this odd because the navigation bar is themed this way without any issues.  `<item name="android:navigationBarColor" tools:targetApi="o_mr1">@color/navigation_bar</item>` [here](https://github.com/AndBible/and-bible/blob/develop/app/src/main/res/values/barstyles.xml#L41)

Another possible issue: `android:windowLayoutInDisplayCutoutMode`, is only available from Android Oreo, so there might be problems with backwards compatibility? (idk)

timbze:
> Another possible issue: `android:windowLayoutInDisplayCutoutMode`, is only available from Android Oreo, so there might be problems with backwards compatibility? (idk)

Yes it will be an issue. You can click the "Override Resource..." link to fix the issue and add the line in v27/barstyles instead. Or do it manually. Edit: actually clicking the link makes Android Studio add a lot more than just the 1 line, so you'll probably have to fix it manually.
Screenshot from 2021-12-13 11-09-40

actuday6418:
I moved the line in question to a new file `v27/barstyles.xml`. Should be backwards compatible now. Does this look right? Copying just the one line makes the application crash for some reason (probably obvious). So I copied it's whole `style` block.

tuomas2:
Merged this, but also improved implementation in 4022c7f (please check!). Was not necessary to copy whole style, but use basetheme instead.

PR #2004, opened 2021-12-26
Save the last search term and added a button to clear text

agrogers:
- The previous search text is now saved
- Button added to clear that text if when you don't want it

image


tuomas2:
Contains new strings, so 4.1

tuomas2:
Bug: 
1. Open Find
2. Write "Jesus"
3. Enable "Remember search text"
4. Click "Find"
5. Click back btn
-> "Remeber search text" is disabled
(should be enabled)

tuomas2:
I think this would be better: https://github.com/AndBible/and-bible/issues/186#issuecomment-451738269
What do you think about implementing that instead?

agrogers:
Yeah, I have wanted that at times also. I think there is a place for both
remembering, clearing with the X and selecting from a history list.

So saving the last term and having an option to disable it is still helpful
to me.

What control would you use for the drop down list?

tuomas2:
> What control would you use for the drop down list?

I have used AutoCompleteTextView in document selection (see document_selection.xml). I think it would work here.

PR #2006, opened 2021-12-27
Remember the last document type selected in the Documents list window

agrogers:
It has frustrated me when i go to the document selector that it always shows me every document. Now it remembers the previous selection.


agrogers:
@tuomas2 , this fails a test that is not related to my code:
net.bible.android.control.page.window.WindowSynchronisationTest > testWindowSyncInMaxMode

timbze:
@agrogers tests fail on develop branch, there is something that needs to be fixed. PRs will have same issue until that gets fixed

agrogers:
> @agrogers tests fail on develop branch, there is something that needs to be fixed. PRs will have same issue until that gets fixed

Thanks Tim. I thought that must have been the case.

tuomas2:
Looks ok to me, merging

PR #2007, opened 2021-12-28
Allow custom actionbar colors for each workspace

agrogers:
This adds two new options to the Color settings page that allows the user to specifiy a color of the workspace header in both day and night modes. 

There are some things I don't  know how to do so if someone wants to give me hand that would be great:
- I need to hide the workspace color option when displaying a the color settings for a window. I believe i know where to do this but I don't know how to access the setting to hide it.
- Only the main bible window view changes the title bar color. The other views keep the original black color. These views seems to use the actionbar while the main bible view uses a toolbar. I think it would be good to have the same color header for all views that are related to a workspace (some app settings are global for example)

And there is some install weirdness:
- It runs fine in the emulator
- When i install directly through Android Studio onto my POCO it works fine
- When i create an APK and try and install on my POCO MIUI device it starts but crashes. 
- When i install onto an Android 8 tablet it installs fine but i get a white actionbar (which seems like the same problem others have been getting)

image

workspace_color


agrogers:
This will need to be meshed with PR #1980 so the status bar color changes as required as well.

agrogers:
And if it is included for release then a database migration addition will be required.

tuomas2:
Interesting! Maybe coloring tint in workspace selector too?

tuomas2:
Tested this now a bit. 

- [x] Wouldn't it be nice to have some indicator color also on Select workspace view?
- [x] Bug (?): New workspaces should have default colors, not current color
- [ ] Default colors should match with current style
- [ ] ~This should not be stored in TextDisplaySettings but rather in WorkspaceSettings. This is not something to be applied to windows, but only to workspaces. Now the color selector is displayed for window settings too.~ maybe we can live with this too. Will think again.
- [ ] Apply color to system top bar too

agrogers:
> Interesting! Maybe coloring tint in workspace selector too?

Definitely. And I might even know how to do that now. I will have a look at it sometime soonish.

I have other things I want to color too - not sure you will like that so much though 🙄

tuomas2 (changes requested):
Needs to be improved before merging. Left tasks in another comment.

tuomas2:
Pushed some commits in this branch, please pull before continuing working on this :-)

tuomas2:
Changed this PR back to draft status now. 

Please change to "ready for review" status when it's ready for review.

agrogers:
@tuomas2 , I think that works nicely.
image

agrogers:
It seems that when Settings are changed via the workspace list those changes are not saved unless you manually push the Save button or click on a workspace. If you push the back button the changes are lost without notification. I don't know how to hook into  the back button.

agrogers:
@tuomas2, I have done the best i can on this now. Will need assistance to take it much further. Comments in the code and I sent you an email as well. Thanks

app/src/main/java/net/bible/service/db/DatabaseContainer.kt
```diff
@@ -981,7 +981,21 @@ private val MIGRATION_56_57_breaklines_in_notes = object : Migration(56, 57) {
         }
     }
 }
-
+private val MIGRATION_57_58_workspace_colors = object : Migration(57, 58) {
+    //TODO: ??? Techincally the workspace colors should not be in 'text_display_settings'. But I fear moving it out will make complicate things a bit.
+    //      At present all the code that manages the saving and copying and initialising of colors also works well for the new workspace colors
+    //      But if I move these settings out somewhere else i am not sure if i will then have to write code to manage that aspect a well. If we do,
+    //      I am not sure that the gain in 'purity' is worth the extra work and potential maitenance. Plus i am not sure how to do it :)
```
tuomas2: > I am not sure that the gain in 'purity' is worth the extra work and potential maitenance.

I am feeling that you migth be right here.

app/src/main/java/net/bible/service/db/DatabaseContainer.kt
```diff
@@ -981,7 +981,21 @@ private val MIGRATION_56_57_breaklines_in_notes = object : Migration(56, 57) {
...
+    override fun doMigrate(db: SupportSQLiteDatabase) {
+        db.apply {
+            val colDefs = "`text_display_settings_colors_dayWorkspaceColor` INTEGER DEFAULT NULL, `text_display_settings_colors_nightWorkspaceColor` INTEGER DEFAULT NULL".split(",")
+            colDefs.forEach {
+                execSQL("ALTER TABLE `Workspace` ADD COLUMN $it")
+                execSQL("ALTER TABLE `PageManager` ADD COLUMN $it")
```
tuomas2, line 997: What if PageManager does not have these fields? Will things break too much?

tuomas2: Hmm it looks like they need to be there for room to work..

agrogers:
I never tried without them but I would assume it would create troubles.

agrogers:
I have now got the Status Bar to have the same color as the toolbar. However there is a bug that I cant fix although I think it is an easy one for you. The status bar color does not update immediately when changing workspaces via the menu. It seems to be setting the statusbar color for the menu activity and not the bible window activity. It works fine if i swipe between workspaces.

As a side note, it doesnt look like it works on my Mi phone because I had 'Hide Notch' enabled. This forces the status bar to be black.

image
image

tuomas2:
@agrogers changed storing workspaceColor to WorkspaceSettings after all.

agrogers:
You are a purist 😉

tuomas2:
It was motivated by fixing #2384 which was caused by this.

agrogers:
Interesting. I have been using my code since I published it and didn't see
this. Well done on fixing it!

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

PR #2019, opened 2021-12-30
Add extra search options to the Find verse dialog

agrogers:
This is still in progress. I would be happy to get other's thoughts.

- 'All endings' adds the wildard * to the end of each search word. Found words are highlighted.
- 'Remember search' remembers what you typed into the search text field
- 'Fuzzy Search' uses the Lucene feature to match words that are similar to the one entered. The example below matches 11 words in the KJV although it is clearly spelt wrong (correct Melchizedek). Unfortunately there is no way at present to highlight the words that matched the search. We need some extra features added to the Lucene module we use to make that happen.
- 'Proximity' will find words that are within X words of each other. Unfortunately this doesn't work across verses and I can't see that happening anytime soon. 

I would like to add a few more options like

- Limiting searches to the words of Jesus or the Gospels or headings
- Find verses where the words are within X number of verses

Does anyone have other suggestions? I should go look through the exising issues I guess...

image


agrogers:
@tuomas2 I am not sure how remembering the previous search state might break other areas. I have tried to be careful and have done a little testing. If there are there are areas that are not immediately obvious that I should test please let me know.

DonaldSch:
What about editing of fuzzy proportion and proximity number?

DonaldSch:
I would prefer an editable search range than a fixed number of range options

DonaldSch:
I think it would be important to leave the option to enter complex logic such as- "+Worship* -(strong:G4352 OR strong:H7812)"

agrogers:
> What about editing of fuzzy proportion and proximity number?

Thanks for the feedback. These options are editable after they are selected.





They can't be used together unfortunately because of limitations on how the search is implemented in the old version of the api we use (if I recall correctly)

agrogers:
> I would prefer an editable search range than a fixed number of range options

I think this will cause problems now but it should be easy to add. You can see the search criteria by enabling 'Show raw search text'. I added this so users could see how to build their own search queries.

But it is cumbersome to have to type it in. Better to have a gui to do it.

agrogers:
> I think it would be important to leave the option to enter complex logic such as- "+Worship* -(strong:G4352 OR strong:H7812)"

This can be added easily.

PR #2058, opened 2022-01-19
Label list filter

### LINKED ISSUE #1620: Proposal for a generic filter option for any list
My lists are starting to get very long. I am usually scrolling for seconds to get to the item I want. This applies to Study Pads, Labels, Dictionary entries, bookmarks, label settings etc

What I was originally thinking of was a simple alphabetical chooser. It shows 'A B C D E F etc' in a row at the top which would filter my lists by first letter. This would get me to most of the things i want with one click and no scrolling. But I have no idea how to implement that for other languages. And it is somewhat wasteful to have, say, X and Z in the list which are rarely used.

The proposal below would work for all languages and be easily customisable:

- A search icon is shown at the top of the list we want to filter
- Pushing it shows a search text box. 
- The text box works like existing filter text boxes - It immediately shows the filtered list below. The difference is...
- Whatever the user types gets saved in the recently searched list. For me i would just type one letter - 'G' say. Then when I return to the form the letter 'G' will appear in the list of search terms. I could, if i wanted, add all the letters in just a minute or so. Then, I can easily filter on the first letter. This should work for any language.
- But i can also add words or multiple letter groups.
- Long pressing the filter option would remove that option from the list. So it would be easy to keep the list clean
- I have shown a single line of saved filter options. But it could be two rows or more deep.
- The same or custom saved lists could be used on all lists that need filtering
- I would also be tempted to remember what was searched for last time and the filtered list shown when returning to it. If that was done we might need an X option to clear the filter.

image

agrogers:
I have this mostly working now. But the code feels really messy, even by my pretty low standards. I will come back to it later and see if i can improve it. If anyone has some helpful suggestions on either interface or code, please let me know. This PR should close https://github.com/AndBible/and-bible/issues/1620

Here is how it works:
- It starts with an 'ALL' button and a magnifing search button
- Pushing the search button toggles the visibiluty of the search textbox and buttons
- The left side button allows you to choose to match either on the start of the label or on any part inside the label (shown as a blue button)
- The Save button will save the text as a button in the row at the top
- It will remember the type of search and display the 'match any part' of the label button with a blue border
- Long-pressing a search button will delete the button
-  The filter seems to work nicely on all the dialogs that use a label list

image

image


agrogers:
Button text shows fine now in dark mode. Options are serialized and saved correctly. I believe this is ready for release now. I don't plan to do anymore on it for a bit. And might I say I couldn't live without it now :)

image

tuomas2:
Then push that "Ready for review" button :-)

agrogers:
> Then push that "Ready for review" button :-)

I should merge the latest 'develop' first. I will do that soon.

agrogers:
OK, develop has been merged.

tuomas2:
- [ ] save button looks like toggle button (confusing)
- [ ] 3rd mode: "regex" to ab-button? and explain in help
- [x] Other modes to ignore regex.
- [x] filtering finds always "Unlabeled"
- [x] Help screen (?-button) should instruct to use filter (for example long-pressing removes saved items, which can't be easily guessed)
- [ ] layout does not work in RTL 
- [x] wont fix ~perhaps all-button is not needed. To disable filter, just tap button again. Or replace with x-button on the right? search button (magnifier glass) should probably be the leftmost button.~
- [x] ~search row should not be visible by default ever~ I'm okay with leaving it as saved setting
- [x] ~check: using fromDb = true in updateLabelList will reload labels from db, i.e. changes will be lost?~ apparently not so, but code needs to be clarified still (fromDb is misleading)
- [x] perhaps show `*`-markings in search buttons (like `cat* ` or `*cat*` etc). I don't think blue border color is very clear indicator.

app/src/main/java/net/bible/android/view/activity/bookmark/ManageLabelItemAdapter.kt
```diff
@@ -72,10 +72,6 @@ class ManageLabelItemAdapter(context: Context?,
                 if (data.showCheckboxes) {
                     checkbox.setOnCheckedChangeListener { _, isChecked ->
                         if (isChecked) {
-                            if (!manageLabels.selectMultiple) {
-                                data.selectedLabels.clear()
-                                data.bookmarkPrimaryLabel = null
-                            }
```
tuomas2, line 78: why this change?

app/src/main/res/values/strings.xml
```diff
@@ -406,6 +406,8 @@
     <string name="auto_assign_labels_primary">Auto-assign as primary</string>
     <string name="label_color">Label color</string>  <!-- can be removed -->
     <string name="label_color_example">Example</string>
+    <string name="match_any_text">*ab*</string>  <!-- Probably does not need to be translated. They are symbols for a button image -->
+    <string name="match_start_of_text">Ab*</string>  <!-- Probably does not need to be translated. They are symbols for a button image -->
```
tuomas2: -> untranslated_strings.xml

tuomas2: though, we have also rtl languages, that should be taken care of. Perhaps by keeping then translatable....

agrogers: I don't think of these as strings but as an icon. I am not sure that there would be a translatable phrase for them. The important character is the *

app/src/main/java/net/bible/android/view/activity/bookmark/ManageLabelItemAdapter.kt
agrogers: I think this was a remnant of when you could choose to select either single or multiple labels. Didn't you hide that option and then no one complained about it so now we can only choose multiple labels. Thus that code is no longer needed.

agrogers:
> * [ ]  save button looks like toggle button (confusing)

Feel free to change it.

> * [ ]  filtering finds always "Unlabeled"

Not on my phone. Can't recall what it did with a default install.

app/src/main/res/values/strings.xml
tuomas2: The issue is that on RTL languages, * should be on the other side than what we think.

tuomas2:
> Not on my phone. Can't recall what it did with a default install.

It is shown in some views, and not others. See for example "label settings" view. 

PS. I'm trying to work on this PR today, let's see how far I get.

app/src/main/java/net/bible/android/view/activity/bookmark/ManageLabels.kt
```diff
@@ -79,15 +99,203 @@ class ManageLabels : ListActivityBase() {
...
+            val orderedList = list
+            orderedList.sortWith(compareBy(String.CASE_INSENSITIVE_ORDER, { it.id }))
+            for (searchOption in orderedList) {
+                val newButton = Button(context)
+                newButton.id = View.generateViewId()
+                newButton.text = searchOption.text
+                newButton.tag = searchOption.id
+                newButton.setBackgroundResource(R.drawable.button_filter)
+                newButton.layoutParams = ViewGroup.LayoutParams(LinearLayout.LayoutParams.WRAP_CONTENT, 60)
+                newButton.minWidth = 80
```
tuomas2: - [x] this and other definitions are pixels if set from code. Should be dp or dip. I think they are not equal.

app/src/main/java/net/bible/android/view/activity/bookmark/ManageLabels.kt
```diff
@@ -538,21 +783,40 @@ class ManageLabels : ListActivityBase() {
...
+        if (fromDb) {
+            shownLabels.clear()
+            Log.i(TAG, "Parsing filter: $filterText")
+            // Check if it is a special type of filter (ie A-C style)
+            if (specialRegex.containsMatchIn(filterText)) {
+                filterText = specialRegex.find(filterText)!!.value
+                filterText = "^[" + filterText.takeLast(filterText.length-1) + "]"
+            }
+            // TODO: too wide catch!
+            val regex = try {filterText.toRegex(RegexOption.IGNORE_CASE)} catch (e:Exception) {"".toRegex()}
```
tuomas2: - [x] catch too wide

app/src/main/java/net/bible/android/view/activity/bookmark/ManageLabels.kt
```diff
@@ -581,11 +845,15 @@ class ManageLabels : ListActivityBase() {
...
-        data.autoAssignLabels.myRemoveIf { !labelIds.contains(it) }
-        data.favouriteLabels.myRemoveIf { !labelIds.contains(it) }
-        data.selectedLabels.myRemoveIf { !labelIds.contains(it) }
+        if (filterText == "") {
 
+            // TODO: changes made in the label dialog are lost when the filter is applied
+            // Some sanity check. These checks clear the changed state of the labels.
+            // AGR: I don't know whether these are really needed. But having them clear these settings when the filter is applied causes me trouble.
+//            data.autoAssignLabels.myRemoveIf { !labelIds.contains(it) }
+//            data.favouriteLabels.myRemoveIf { !labelIds.contains(it) }
```
tuomas2: - [x] TODO

agrogers:
I have added the HideKeyboard flag back in because I believe it needs it. Without it the keybaord disappears after each keystroke - both on the emulator and my device.

app/src/main/java/net/bible/android/view/activity/bookmark/ManageLabels.kt
```diff
@@ -70,6 +87,23 @@ fun WorkspaceEntities.WorkspaceSettings.updateFrom(resultData: ManageLabels.Mana
...
+class SearchOption(
+    var text: String,
+    var isSearchInsideText: Boolean,
+    var id: String = "",
+    var regex: String = "",
+    var name: String = ""
+) {
+    // TODO: Presently 'name' is not used. I want to show a dialog on the long press of a button
+    //  that allows the deleting or renaming of a button. This allows for complex regex strings but with a more human readable name
+    init {
```
tuomas2: Removing name for now. It can be added in the future if functionality is added.

agrogers:
Last pullrequest removes the wasted space due to the almost always empty column for the Primary icon.

image

app/src/main/java/net/bible/android/view/activity/bookmark/ManageLabels.kt
```diff
@@ -79,15 +115,184 @@ class ManageLabels : ListActivityBase() {
...
+    private lateinit var lastSelectedQuickSearchButton: Button
+    private var showTextSearch = false
+    private var searchInsideTextButtonActive = false
+
+    private val searchOptionList: ArrayList<SearchOption> = ArrayList()
+    private fun findSearchOptionListItem(id: String): SearchOption? = searchOptionList.find { it.id == id }
+
+    private fun loadFilteringSettings() {
+        searchInsideTextButtonActive = CommonUtils.settings.getBoolean("labels_list_filter_searchInsideTextButtonActive", false)
+        showTextSearch = CommonUtils.settings.getBoolean("labels_list_filter_showTextSearch", false)
```
tuomas2, line 130: Do we want that this setting is remembered? I would assume people do not want to have this on, unless they tap magnifier glass icon.

app/src/main/java/net/bible/android/view/activity/bookmark/ManageLabels.kt
```diff
@@ -79,15 +115,184 @@ class ManageLabels : ListActivityBase() {
...
+        val options = CommonUtils.settings.getString("labels_list_filter_searchTextOptions") ?: ""
+        searchOptionList.clear()
+        if (options.isNotEmpty() && options.first() == '[') {
```
tuomas2: why the second condition?

app/src/main/java/net/bible/android/view/activity/bookmark/ManageLabels.kt
```diff
@@ -538,21 +747,40 @@ class ManageLabels : ListActivityBase() {
...
+            if (specialRegex.containsMatchIn(filterText)) {
+                filterText = specialRegex.find(filterText)!!.value
+                filterText = "^[" + filterText.takeLast(filterText.length-1) + "]"
+            }
+            // TODO: too wide catch!
+            val regex = try {filterText.toRegex(RegexOption.IGNORE_CASE)} catch (e:Exception) {"".toRegex()}
+            // Exclude the 'unlabeled label' but include all selected labels plus those that match the filter clause.
+            // We always want to see the labels we have selected even when filtering.
+            shownLabels.addAll(allLabels.filter {
+                !it.isUnlabeledLabel and (
```
tuomas2: `and` and `or` are bitwise operators in kotlin. Use && and ||.

tuomas2:
> [ ] 3rd mode: "regex" to ab-button? and explain in help

What do you think, would full 'regex mode' be needed? Would be used by geeks only, of course... Now all regex will be escaped, so patterns can't be used (except those simple ones provided by `*Ab*/Ab*` button).

tuomas2:
> Last pullrequest removes the wasted space due to the almost always empty column for the Primary icon.
> 
> image

~Side effect is that now it is more difficult to change primary label. (can't be done from labels listing)~ no, it is possible (because all selected will have it visible)

tuomas2:
RTL now:

image

app/src/main/java/net/bible/android/view/activity/bookmark/ManageLabels.kt
tuomas2: improved

app/src/main/java/net/bible/android/view/activity/bookmark/ManageLabels.kt
tuomas2: fixed

agrogers:
> > [ ] 3rd mode: "regex" to ab-button? and explain in help
> 
> What do you think, would full 'regex mode' be needed? Would be used by geeks only, of course... Now all regex will be escaped, so patterns can't be used (except those simple ones provided by `*Ab*/Ab*` button).
> 
> 

If I could name an entry I would use regex. But I agree only feels would use it. Once it was practical I thought I would add some regex strings I thought popular to the manual or popup help and then people just had to copy and paste.

agrogers:
Does the modifications show the * for all modes or just the new mode? I used the color coding of buttons because I was trying to keep the buttons as small as possible. I could almost get A through Z on two lines.

app/src/main/java/net/bible/android/view/activity/bookmark/ManageLabels.kt
agrogers: I almost always have the text search enabled. I find that just as fast or faster than using the buttons :) Should have just done that in the first place!

agrogers: I almost always have the text search enabled. I find that just as fast or faster than using the buttons :) Should have just done that in the first place!

tuomas2:
> Does the modifications show the * for all modes or just the new mode? I used the color coding of buttons because I was trying to keep the buttons as small as possible. I could almost get A through Z on two lines.

I think color coding does not communicate user what is going on. I prefer wasting that space because user understands it better.

agrogers:
That's OK. I can change it back on my build if I don't like it 🙂

But I think it is wasted space - repeating the same thing 26 times in
limited space is usually a indicator to me that there is a better way of
doing it. I would put this in the category of 'let the interface teach the
user' rather than spelling it out for them.

tuomas2:
I think this is now good (enough) to go. Let's see if beta testing will give reasons for improving this further. 

Very convenient and nice feature! Thank you.

PR #2061, opened 2022-01-19
Search results with statistics

### LINKED ISSUE #2049: New search statistics display proposal
I have been working on a new search results window. It is mostly working but has more than a few problems which I might not be able to resolve by myself. I thought I would post my progress here to get some feedback about what others might find useful.

What currently works is:

- Search hits now highlight correctly. That is, whole words are highlighted even when searching using the wildcard *
- The results are broken down into _By Book_ and _By Word_ totals
- Clicking the _By Book_ book button changes to the Results list and scrolls to the selected book
- _By Word_ summary is presently based on the highlighted words. That makes sense sometimes but not so much for Strongs searches because, as can be seen, the highlighted words have a lot extra unneeded words.

To do:

- The _By Word_ button is supposed to filter the search results.
- Cleanup the strongs words to try and get a more useful list (eg merge, 'will we prevail', 'prevailed', 'have prevailed' into the single word 'prevail'
- Add more statistics?

image

image

image

PS: @tuomas2 , my recent comments about the need for a search list as well as a pseudo document search list is, I think, strengthened by this proposal. That is, having a standard dialog to show the results in also allows us to show stats like these. But perhaps, we could provide an option on the search form itself that  allows the user to choose what search display he/she wants to see.

tuomas2:
Closes #2049


app/src/main/AndroidManifest.xml
```diff
@@ -94,6 +94,11 @@
             android:configChanges="keyboardHidden|orientation|screenSize|locale"
             android:label="@string/search"
             />
+        <activity
+            android:name="net.bible.android.view.activity.search.MySearchResults"
```
tuomas2, line 123: Isn't this temporary (2 search results activities)

tuomas2:
- [ ] Search activity does not have defaults set (Whole Bible, All words should be set by default)

app/src/main/java/net/bible/android/control/search/SearchControl.kt
```diff
@@ -142,6 +167,26 @@ class SearchControl @Inject constructor(
         return searchResults
     }
 
+    fun getSearchResultVerseText(key: Key?): String {
```
tuomas2, line 170: unused function

app/src/main/java/net/bible/android/control/search/SearchControl.kt
```diff
@@ -232,7 +277,7 @@ class SearchControl @Inject constructor(
         const val TARGET_DOCUMENT = "TargetDocument"
         private const val STRONG_COLON_STRING = LuceneIndex.FIELD_STRONG + ":"
         private const val STRONG_COLON_STRING_PLACE_HOLDER = LuceneIndex.FIELD_STRONG + "COLON"
-        const val MAX_SEARCH_RESULTS = 1000
+        const val MAX_SEARCH_RESULTS = 10000
```
tuomas2, line 280: Quite a lot?

app/src/main/java/net/bible/android/view/activity/navigation/GridChoosePassageBook.kt
```diff
@@ -324,16 +324,43 @@ class GridChoosePassageBook : CustomTitlebarActivityBase(R.menu.choose_passage_b
...
+        private val PAULINE_COLOR = Color.rgb(0xF5, 0xf5, 0x21)
         private val GENERAL_EPISTLES_COLOR = Color.rgb(0x67, 0xCC, 0x66) // changed 99 to CC to make a little clearer on dark background
         private val REVELATION_COLOR = Color.rgb(0xFE, 0x33, 0xFF)
         private val OTHER_COLOR = ACTS_COLOR
 
         public const val BOOK_GRID_FLOW_PREFS = "book_grid_ltr"
         public const val BOOK_GRID_FLOW_PREFS_GROUP_BY_CATEGORY = "book_grid_group_by_category"
         private const val TAG = "GridChoosePassageBook"
+
+        fun getBookTextColor(bookNo: Int): Int {
```
tuomas2: Why here? Not used in GridChoosePassageBook activity.

Perhaps need to move all these color-related code somewhere else, if they are used in multiple places.

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
```diff
@@ -0,0 +1,438 @@
...
+    resources.getString(R.string.add_bookmark_whole_verse1),
+    resources.getString(R.string.by_book),
+    resources.getString(R.string.by_word)
+)
+var mCurrentlyDisplayedSearchResults: List<Key> = ArrayList()  // I tried to make this a non-global but i had the same problem as before where the values in the variable was one behind the search. I don't know why!!
+const val verseTabPosition = 0
+private const val bookTabPosition = 1
+private const val wordTabPosition = 2
+
+class MyTimer(val name:String){
```
tuomas2: - [x] remove debug code before merge

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
```diff
@@ -0,0 +1,438 @@
...
+    }
+    fun reset() {totalTime = 0}
+    fun log() {
+        Log.e("MyTimer", name + ": " + (totalTime/1000000) + "mS\n")
+    }
+}
+
+class BookStat(val book: String, var count: Int,
+               val bookInitials: String,
+               val bookOrdinal:Int,
```
tuomas2, line 64: not used

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
```diff
@@ -0,0 +1,438 @@
...
+                val color: Int) {
+    override fun toString(): String = "$book: $count"
+}
+class WordStat(val word: String,
+               var verseIndexes: IntArray,
+               val originalWord: String) {
+    override fun toString(): String = "$originalWord: ${verseIndexes.count()}"
+}
+
+private var wordHits: ArrayList<Pair<String, ArrayList<Int>>> = arrayListOf()
```
tuomas2, line 76: not used

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
```diff
@@ -0,0 +1,438 @@
...
+    @JvmField
+    var osisKey: String?
+    @JvmField
+    var reference: String?
+    @JvmField
+    var translation: String?
+    @JvmField
+    var verse: String?
+    @JvmField
+    var verseHtml: String?  // I would prefer to pass a spannable string but i dont know how to make it parcelable
```
tuomas2: - [ ] TODO

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
```diff
@@ -0,0 +1,438 @@
...
+
+    /* What I really want to do is make the 'Key' parcelable but I don't know how to do that. So instead I have to send the properties I need and get the key later on */
```
tuomas2, line 86: - [ ] TODO

tuomas2: Key has been serialized elsewhere. Need to check if we can use same tricks here.

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
```diff
@@ -0,0 +1,438 @@
...
+        Log.i(TAG, "Displaying Search results view")
+        totalTime.reset()
+
+        binding = SearchResultsStatisticsBinding.inflate(layoutInflater)
+        setContentView(binding.root)
+
+        buildActivityComponent().inject(this)
+
+        val bundle = Bundle()
+        bundle.putString("edttext", "From Activity")
```
tuomas2, line 170: ?

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
```diff
@@ -0,0 +1,438 @@
...
+
+class SearchResultsPagerAdapter(private val context: Context, fm: FragmentManager,
+                                searchControl: SearchControl,
+                                activeWindowPageManagerProvider: ActiveWindowPageManagerProvider,
+                                intent: Intent,
+                                val mSearchResultsArray:ArrayList<SearchResultsData>,
+                                val bookStatistics: MutableList<BookStat>,
+                                val wordStatistics: MutableList<WordStat>
+) :
+    FragmentPagerAdapter(fm) {
```
tuomas2, line 390: Deprecated
Switch to androidx.viewpager2.widget.ViewPager2 and use androidx.viewpager2.adapter.FragmentStateAdapter instead.

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
```diff
@@ -0,0 +1,438 @@
...
+    }
+}
+
+class SearchResultsPagerAdapter(private val context: Context, fm: FragmentManager,
+                                searchControl: SearchControl,
+                                activeWindowPageManagerProvider: ActiveWindowPageManagerProvider,
+                                intent: Intent,
+                                val mSearchResultsArray:ArrayList<SearchResultsData>,
+                                val bookStatistics: MutableList<BookStat>,
+                                val wordStatistics: MutableList<WordStat>
```
tuomas2: we could pass activity as context and avoid long list of parameters

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
```diff
@@ -0,0 +1,438 @@
...
+        viewPager.offscreenPageLimit = 2    // The progressbar on the 3rd tab goes to zero when the view is lost. So just keep it in memory and all is fine. It is not a big view so i think it is ok.
+        val tabs: TabLayout = binding.tabs
+        tabs.setupWithViewPager(viewPager)
+
+        searchResultsActionBarManager.registerScriptureToggleClickListener(scriptureToggleClickListener)
+        setActionBarManager(searchResultsActionBarManager)
+        supportActionBar?.setDisplayHomeAsUpEnabled(true)
+        isScriptureResultsCurrentlyShown = searchControl.isCurrentlyShowingScripture
+
+        GlobalScope.launch {
```
tuomas2, line 190: TODO

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
```diff
@@ -0,0 +1,438 @@
...
+            prepareResults()
+        }
+    }
+
+    private suspend fun prepareResults() {
```
tuomas2: most of stuff in Main, only fetching in IO

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
```diff
@@ -0,0 +1,438 @@
...
+                getString(R.string.search_showing_first, SearchControl.MAX_SEARCH_RESULTS)
+            } else {
+                getString(R.string.search_result_count, mSearchResultsHolder!!.size())
+            }
+            withContext(Dispatchers.Main) {
+                Toast.makeText(this@MySearchResults, msg, Toast.LENGTH_SHORT).show()
+            }
+            isOk = true
+            fetchResultsTimer.stop(true)
+        } catch (e: Exception) {
```
tuomas2, line 228: too wide catch

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
```diff
@@ -0,0 +1,438 @@
...
+            }
+            mSearchResultsHolder = searchControl.getSearchResults(searchDocument, searchText)
+            // tell user how many results were returned
+            val msg:String = if (mCurrentlyDisplayedSearchResults.size >= SearchControl.MAX_SEARCH_RESULTS) {
+                getString(R.string.search_showing_first, SearchControl.MAX_SEARCH_RESULTS)
+            } else {
+                getString(R.string.search_result_count, mSearchResultsHolder!!.size())
+            }
+            withContext(Dispatchers.Main) {
+                Toast.makeText(this@MySearchResults, msg, Toast.LENGTH_SHORT).show()
```
tuomas2: can use ToastEvent

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
```diff
@@ -0,0 +1,438 @@
...
+            verseTextSpannableTimer.stop()
+
+            mSearchResultsArray.add(SearchResultsData(listIndex, key.osisID.toString(), key.name,searchDocument, "text", verseTextSpannable!!.toHtml()))
+
+            // Add book to the book statistics array
+            bookStatsTimer.start()
+            val bookOrdinal = ((key as Verse).book as BibleBook).ordinal
+            var mBibleBook = BibleBook.values()[bookOrdinal]
+            var bookNameLong = versification.getLongName(mBibleBook)  // key.rootName
+            val bookStat = bookStatistics.firstOrNull{it.book == bookNameLong}
```
tuomas2: - [ ] It looks like bookStatistics and other statistics containers could be maps, which is more efficient

agrogers:
This code needs plenty of love (I thought the label filtering code didn't need much work but you did a lot on it... so I shudder to think what there is for this). I don't have any problems with the code - it has been bug free for me over the last 6 months. Things to do:

- [ ] Do the search in a different thread. At present it locks the interface so long searches are too slow.
- [ ] Display the results progressively. The first results can be shown almost immediately because the search itself is fast. It is the retrieval of the entire verse and the parsing of those verses that is slow.
- [ ] Maybe find a better way to parse the results. Parsing is needed to work out the word stats. That is, it needs to look at the text of every verse, search for the Strongs number (or word), work out what words it applies to, create a list of those words and count their uses etc
- [ ] The 'Verses' and 'Books' totals can be displayed immediately but the 'Words' total must be delayed until all the verses have been parsed. So indicate that the Word total is incomplete (perhaps '1+' which updates as the words are found)
- [ ] 'Show key words only' is not implemented. A strongs search will often show a LOT of word variations. My idea was to consolidate those words as much as possible. The plan was to look through the list for all single words longer than 3 characters. Then, if that single word is used elsewhere, collapse the other hits into that word. For example, H7641 gives 'among the ears of grain, branches, ears, ears of grain, of grain, the ears'. My approach would reduce the results to 'ears, branches, of grain' which is more useful from a statistics perspective. Some strongs search results give a crazy long list of words. I feel this would give reasonable results 90% of the time.

tuomas2 (commented):
First batch of comments

app/src/main/java/net/bible/android/control/search/SearchControl.kt
agrogers: I dont think we need to limit it at all. The most hits i got was about 4000 for some common words. The search itself is very fast. Even big searches in its current form were acceptable (but slow) for me. The complete search results are valuable now because of the stats. My suggestions it make it very large.

app/src/main/java/net/bible/android/control/search/SearchControl.kt
agrogers: Quite possible.

app/src/main/java/net/bible/android/view/activity/navigation/GridChoosePassageBook.kt
agrogers: I cant recall what i did here. But yes, they are now used in two places and i have other places i would like to use them as well. I think i wrote some functions to get what i wanted and added them somewhere. And i seem to recall not knowing where the best place to put them would be.

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
agrogers: I think I wanted to do this to try and improve performance. Not sure it saves much time though.

app/src/main/AndroidManifest.xml
agrogers: Maybe. I don't understand much about the manifest :)

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
agrogers: I agree!

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
agrogers: This was my first attempt at pages and fragments and other stuff. So there is a lot in here that needs fixing or improving. I think it might almost be better for someone who knows what they are doing just to start again.

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
agrogers: Yeah, I was using some outdated stuff because the sample code from elsewhere and inside AB was using the old style.

app/src/main/AndroidManifest.xml
tuomas2: I mean, we have now SearchResults, and MySearchResults. I think SearchResults is deprecated by MySearchResults. End result should be that we remove old SearchResults and rename MySearchResults -> SearchResults.

agrogers:
Yes, I wanted to keep the old way so i could easily go back and look
at how it worked. But it is not needed.... except if someone hates the new
search and just wants the simple old search results back. If the results
with Statistics works as fast as the old search then there is no need for
both.

tuomas2:
I took new strings from here to develop already, so that this can be included easily even after 4.1 release if this is not ready before that.

I will leave this now waiting for the mentioned improvements (anyone is free to contribute).

app/src/main/AndroidManifest.xml
agrogers: I am sure i had answered this somewhere else. Yes, MySearchResults is a duplicate. The old one could be removed.

app/src/main/java/net/bible/android/view/activity/navigation/GridChoosePassageBook.kt
agrogers: I think you have moved this and i have incorporated that change.

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
agrogers: Done

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
agrogers: I dont know how to run this in the background sorry.

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
agrogers: Not sure what to narrow it to.

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
agrogers: Done i think.

app/src/main/java/net/bible/android/view/activity/search/MySearchResults.kt
agrogers: Outdated?

agrogers:
I have implemented the 'Show key words only' now. Done very little testing on it. Will do that over the next few weeks. It's rudimentary but seems to give good results.
image
image

PR #2072, opened 2022-01-25
add additional info to Dictionary Word picker

RuggMatt:
This just adds some additional info to the Hebrew/Greek word picker 
image

I am not sure if this is the best way to extract the info. It works for the AndBible Greek and Hebrew dictionaries, but nothing else. I am not sure if how I look up the additional info is the best way to go about it, or if there is a more general way to query for the `entryFree` info.

**Screenshots**

### DIFF
--- a/app/src/main/java/net/bible/android/view/activity/navigation/ChooseDictionaryWord.kt
+++ b/app/src/main/java/net/bible/android/view/activity/navigation/ChooseDictionaryWord.kt
@@ -20,13 +20,11 @@ package net.bible.android.view.activity.navigation
 import android.app.Activity
 import android.content.Intent
 import android.os.Bundle
-import android.os.Handler
 import android.text.Editable
 import android.text.TextWatcher
 import android.util.Log
 import android.view.View
 import android.widget.ArrayAdapter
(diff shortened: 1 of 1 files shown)


agrogers:
Screenshot looks good to me. How does it look with normal dictionaries like Smiths or Easton?

RuggMatt:
@agrogers other dictionaries already have the name as the key.
image

Strongs Greek and Hebrew just had the Strongs number. Now it also shows the related word next to it.

agrogers:
Ah, i get it now.

I would also like to show the first few lines of the actual dictionary
entry in the list. There have been more than a few occasions where that
would have been all i needed.

But I think that will really slow down retrieval.

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

RuggMatt:
@agrogers I did some testing and did not find any performance issues with retrieval. 
image

tuomas2:
Looks good to me, merging

tuomas2:
As always, unit testing would be nice, but I'm assuming you'll fix the bugs if there will be any related :-)

Issue #1473, opened 2021-08-02
Import e-sword, mysword & mybible modules


cabo40:
**Is your feature request related to a problem? Please describe.**
It would be great to be able to import e-sword or mysword modules 

**Describe the solution you'd like**
When importing a bible, open the possibility to select an e-sword or mysword file and import it.

**Describe alternatives you've considered**
both e-sword and mysword are sqlite databases with a verse per entry. By taking a peek at the code it seems like a hard feat to implement because of the way documents are stored. Maybe have two ways of reading local bibles?

tuomas2:
It's indeed possible to create other than jsword module backends, that generate OSIS fragments from whatever source.

tuomas2:
But why do we need that actually?

cabo40:
There is a wider catalog of translations in esword format (or at least easier to find)  for spanish language (https://www.bibliotecahispana.com/ , https://esword-espanol.blogspot.com). Also I haven't found a way to convert from esword to sword, and this tool https://github.com/schierlm/BibleMultiConverter can't convert either, but can convert many formats to esword, opening also the possibility to import an even wider catalog of bibles.
What do you think? :)

timbze:
> this tool https://github.com/schierlm/BibleMultiConverter can't convert either

It looks like this tool can "export" to OSIS format. CrossWire has tools to convert from OSIS to SWORD. So maybe there would be a way to convert from e-sword to OSIS, then to SWORD. I don't know if it would work, but maybe.

BryceSteuer:
@tuomas2 
There is a massive and free library of e-sword modules on the web, including but not limited to (study) Bibles, commentaries, sermons, books, et cetera. [This site](http://www.biblesupport.com/) alone has thousands of high-quality modules.

cabo40:
> CrossWire has tools to convert from OSIS to SWORD

If you could point to this tool it would be great! I could import my esword modules to andbible this way, but still hope this can get implemented :pray:.

timbze:
> > CrossWire has tools to convert from OSIS to SWORD
> 
> If you could point to this tool it would be great! I could import my esword modules to andbible this way, but still hope this can get implemented 🙏.

I am not very familiar with the tools. The original source code is on SVN [here](https://crosswire.org/svn/sword/trunk/). Here's, what looks like, a [working mirror on Github](https://github.com/bibletime/crosswire-sword-mirror).

[...]

PR #2088, opened 2022-02-10
Sqlite-based backend (#1473)

tuomas2:
#1473


tuomas2:
Proof of concept works (to some extent, at least!)! 

This was read directly from `gsb1951.bbl.mybible`.

image

PR #2100, opened 2022-02-17
CI: Add app version suffix for -alpha and -beta builds

timbze:
I think this would be nice... @tuomas2 

When tagging for alpha- or beta- builds, this will add -alpha or -beta to the version name before building apk

Side note: I tested `sed` command to work but I have not tested in GitHub CI

### DIFF
--- a/.github/workflows/build-apk.yml
+++ b/.github/workflows/build-apk.yml
@@ -24,7 +24,7 @@ jobs:
       uses: actions/setup-java@v2
       with:
         distribution: 'zulu'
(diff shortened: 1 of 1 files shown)


tuomas2:
Thanks, will check if it works

tuomas2:
It would be nice if we would get some automatic text to alpha build descriptions / titles so that there would not be danger of confusing between prod builds and alphas.

tuomas2:
Though, version number is visible also in release title, so maybe it helps already.

PR #2125, opened 2022-03-11
Fix "Set as current day" in reading plan

Adriankhl:
I believe "Set as current day" should also set the start day of a reading plan such that the current plan day is today. This PR fixes the issue.

I have also removed MaxPermSize from `gradle.properties` since it is deprecated and it produces error in latest JDK.

### DIFF
--- a/app/src/main/java/net/bible/android/view/activity/readingplan/DailyReading.kt
+++ b/app/src/main/java/net/bible/android/view/activity/readingplan/DailyReading.kt
@@ -336,9 +336,17 @@ class DailyReading : CustomTitlebarActivityBase(R.menu.reading_plan) {
             try {
                 Dialogs.instance.showMsg(R.string.msg_set_current_day_reading_plan, true)
                 {
+                    // Change start date so that the current plan day is today
+                    val planStartDate = Calendar.getInstance()
+                    planStartDate.add(Calendar.DATE, - (dayLoaded - 1))
+                    readingPlanControl.setStartDate(readingsDto.readingPlanInfo, planStartDate.time)
+
                     // set previous day as finish, so that today's reading status will not be changed
                     readingPlanControl.done(readingsDto.readingPlanInfo, dayLoaded - 1, true)
+
                     updateTicksAndDone()
+
+                    loadDailyReading(planCodeLoaded, dayLoaded)
                 }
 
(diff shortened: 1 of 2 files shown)


tuomas2:
Should this PR is towards current-stable (i.e. fixing production issue)?

app/src/main/java/net/bible/android/view/activity/readingplan/DailyReading.kt
```diff
@@ -336,9 +336,17 @@ class DailyReading : CustomTitlebarActivityBase(R.menu.reading_plan) {
...
+                    // Change start date so that the current plan day is today
+                    val planStartDate = Calendar.getInstance()
+                    planStartDate.add(Calendar.DATE, - (dayLoaded - 1))
+                    readingPlanControl.setStartDate(readingsDto.readingPlanInfo, planStartDate.time)
+
                     // set previous day as finish, so that today's reading status will not be changed
                     readingPlanControl.done(readingsDto.readingPlanInfo, dayLoaded - 1, true)
+
                     updateTicksAndDone()
+
```
timbze:
```suggestion
```

This is now not needed since you're calling `loadDailyReading`

timbze:
Beside my suggestion this PR looks good. And it's a great idea. Thanks @Adriankhl

timbze:
> Should this PR is towards current-stable (i.e. fixing production issue)?

It's not a big issue per se, but might as well go to current-stable. I realize that changing the branch here now doesn't work (and it makes sense, just didn't think of it).

tuomas2:
> > Should this PR is towards current-stable (i.e. fixing production issue)?
> 
> It's not a big issue per se, but might as well go to current-stable. I realize that changing the branch here now doesn't work (and it makes sense, just didn't think of it).

Yep should cherry-pick commits based on current-stable and then force-push.

tuomas2:
> > > Should this PR is towards current-stable (i.e. fixing production issue)?
> > 
> > 
> > It's not a big issue per se, but might as well go to current-stable. I realize that changing the branch here now doesn't work (and it makes sense, just didn't think of it).
> 
> Yep should cherry-pick commits based on current-stable and then force-push.

Did just that. It's in now! Thank you!

Issue #2133, opened 2022-03-17
MyBible: OT quotes not displayed properly


tuomas2:
> Another issue I came across with the German module "S'00" (Schlachter 2000) is that when Christ quotes the OT in Matthew 5, the quotation appears as a reference and not within the text. The module works fine with MyBible.

https://support.andbible.org/scp/tickets.php?id=1335

timbze:
I have found a few MyBible modules outside of MyBible (i.e. NKJV is one) and it does the same for words that KJV has italicized, meaning that specific word is not in the Greek. Happens for words like "is" etc. I did not report this before because I wasn't sure that the module was 100% good, but I'm suspecting that the issue is along the same lines of the OP.

"w" = Let
"x" = of
"y" = you

Top is KJV SWORD module
Bottom is NKJV MyBible module
![image](https://user-images.githubusercontent.com/35117769/163226061-d95bb934-4152-45c5-b4d9-e99b110973f7.png)

PR #2150, opened 2022-04-20
Changed how <i> tags in text from MyBible modules are handled

uniqueid001:
Display an `<i>` element from the text of a MyBible module like an
OSIS `<transChange type="added">` element.

This fixes the issue described by timbze in https://github.com/AndBible/and-bible/issues/2133#issuecomment-1098249508

### DIFF
--- a/app/bibleview-js/src/components/MyBible/I.vue
+++ b/app/bibleview-js/src/components/MyBible/I.vue
@@ -1,0 +1,41 @@
+<!--
+  - Copyright (c) 2022 Martin Denham, Tuomas Airaksinen and the And Bible contributors.
+  -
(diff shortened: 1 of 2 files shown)


tuomas2:
This should go to current-stable branch. You could may rebase your branch yourself or I can cherry-pick too to current-stable (tomorrow).

uniqueid001:
I fixed the PR.  I committed on current-stable, but didn't select it when I created the PR very late last night.

tuomas2:
simplified a bit ^

uniqueid001:
> simplified a bit ^

I tried it both ways, and then decided to submit it more consistent with the OSIS transChange tag handling.  : )

PR #2157, opened 2022-05-02
Workaround duplicate classes from transitive dependencies

uniqueid001:
Reference: https://github.com/android/android-test/issues/861

The transitive dependencies of espresso-accessibility and
espresso-contrib include versions of both "checker", and "checker-qual"
which contains a portion of "checker".  This results in duplicate
classes which cause the build of "androidTest" variants to fail.

This commit applies the workaround, described in the referenced github
issue, of excluding "checker" from the transitive dependencies.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -342,9 +342,15 @@ dependencies {
 
     // Espresso dependencies
     androidTestImplementation("androidx.test.espresso:espresso-core:3.4.0")
-    androidTestImplementation("androidx.test.espresso:espresso-contrib:3.4.0")
+    androidTestImplementation("androidx.test.espresso:espresso-contrib:3.4.0") {
+        // https://github.com/android/android-test/issues/861#issuecomment-1067448610
+        exclude(group="org.checkerframework", module="checker")
+        }
     androidTestImplementation("androidx.test.espresso:espresso-intents:3.4.0")
-    androidTestImplementation("androidx.test.espresso:espresso-accessibility:3.4.0")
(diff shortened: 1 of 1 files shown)


agrogers:
Thanks for that. I might try another test when I get time and see if it
works again for me. I don't know what my problem is so I I dont know if it
will solve it ☺️

uniqueid001:
The symptom that this PR addresses is below.  It only occurs when building "androidTest" for instrumentation tests.
`
Execution failed for task ':app:checkStandardDebugAndroidTestDuplicateClasses'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.CheckDuplicatesRunnable
   > Duplicate class org.checkerframework.checker.compilermsgs.qual.CompilerMessageKey found in modules jetified-checker-3.1 (org.checkerframework:checker:3.1.1) and jetified-checker-qual-3.13.0 (org.checkerframework:checker-qual:3.13.0)
     Duplicate class org.checkerframework.checker.compilermsgs.qual.CompilerMessageKeyBottom found in modules jetified-checker-3.1 (org.checkerframework:checker:3.1.1) and jetified-checker-qual-3.13.0 (org.checkerframework:checker-qual:3.13.0)
[...]
`

I wonder if I should submit it for the current-stable branch instead of develop?

agrogers:
Thanks for clarifying.

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

tuomas2:
wanted to push this to current-stable so merged manually

PR #2158, opened 2022-05-02
Replace ActivityTestRule with ActivityScenario.launch()

uniqueid001:
ActivityTestRule is deprecated, superseded by ActivityScenario.launch()


tuomas2:
Just as a background:

Android tests (that are run in a real emulator) are kind of a stub for https://github.com/AndBible/and-bible/pull/803 which was a nice idea that I didn't find enough motivation to impelement. So there is nothing really useful at the moment.

Issue #817, opened 2020-08-21
Create search index user friendliness


timbze:
When searching module for the first time, this message comes up: 
**An index is required to search. Create index for KJV?** With option to either **Cancel** or **Create**

I doubt that it's a good idea to automatically create an index for each module automatically after install (e.g. I don't normally search in commentaries).

However, we could make the message more user friendly for people who have no idea what an index is, for example we could have a message something like:
**In order to search KJV, And Bible needs to do a one-time search index generation which may take a few minutes.** With the options to simply "Cancel" or "**Continue**" (i.e. continue instead of create)

And then on "Continue" press, find window would open with generate status at bottom, but find criteria could already be entered. The search button would not become active until index generation is complete.

timbze:
- [ ] When creating index it shows this screen (what is visible has nothing to do with index creation, and it's not scrollable).
![image](https://user-images.githubusercontent.com/35117769/115780614-9cf98a00-a376-11eb-8bb7-83310cbfde29.png)

agrogers:
We have changed the message about an 'index'. I think it is a little clearer.

I suggest we delay this one to after v3.4 release

tuomas2:
> * [ ]  When creating index it shows this screen (what is visible has nothing to do with index creation, and it's not scrollable).

More specifically
- [ ] should clear finished tasks automatically from the list
- [ ] should make list scrollable
- [ ] should position ok button below the list

agrogers:
Is there any reason that these indexes cant be generated whenever a new document is included? How much space do the indexes consume? If it is not much i would just learn the indexing process in the background when the Bible is added.

tuomas2:
Maybe automatic background indexing could be done, optionally (setting to disable this should be there probably too). Notification is displayed to user while indexing takes place.

agrogers:
A toast message and the general Android notifications are already
implemented I think.

___________________________
*www.TenOutOfTen.org* <https://www.TenOutOfTen.org>
***@***.***
(+61) 0477 174 193 (Australia)
(+95) 09 250018669 (Myanmar)

tuomas2:
Yes.

timbze:
> Is there any reason that these indexes cant be generated whenever a new document is included? How much space do the indexes consume? If it is not much i would just learn the indexing process in the background when the Bible is added.

[...]

PR #2159, opened 2022-05-02
Bug/#817 improve layout of progress status

uniqueid001:
#817: wrap the LinearLayout in a ScrollView, and adjust the layout so it does not overlap with the "OK" button.


tuomas2:
Wow nice (props for tests :-) ). Thanks.

PR #2162, opened 2022-05-05
Replace deprecated lintOptions{} block with lint{}

uniqueid001:


tuomas2:
Hmm I think there was some issue which prevented from changing this...

(btw small fixes like this can go to current-stable branch)

tuomas2:
FIxed this is in current-stable branch

PR #2179, opened 2022-06-11
Adjusted Verse selector button color so that it is easy to see what verse is currently selected 

agrogers:
I have always found it hard to see which book or chapter was highlighted indicating the current verse.  This change makes it very easy. I have also enabled the highlight for the current verse - I sometimes want to know that. The button color uses the default color for the book - so it changes depending on which book is current.

image

image

image

image


tuomas2:
Looks good! Thank you!

Issue #1857, opened 2021-10-14
App title should be changed to something <= 30 characters


tuomas2:
Google is changing their policy! App title must be changed to something less than 30 charaters (in all languages).

[Google's updated guidelines can be read here.](https://android-developers.googleblog.com/2021/04/updated-guidance-to-improve-your-app.html)

Not sure, but I guess we might have enough time to do this change towards 4.1.

tuomas2:
> > [!] Google Api Error: Invalid request - The title for the app is too long for language en-US.

Can't update metadata any more! 
So no changes to app descriptions etc until name change is done.

tuomas2:
I think we do not have any option but drop "by And Bible Open Source Project" from our Play Store official name due to this new regulation. Feel free to start discussing ideas for new name. 

Perhaps we should keep doors open to something completely new?

agrogers:
What about "Bible Study App, by And Bible"

___________________________
*www.TenOutOfTen.org* <https://www.TenOutOfTen.org>
***@***.***
(+61) 0477 174 193 (Australia)
(+95) 09 250018669 (Myanmar)

BryceSteuer:
I always have thought "Bible Study App" was a bit generic, but I guess if it's an open source test...

agrogers:
Yeah, I probably agree with you now. We have had a few discussions about
the name over the last year.

One of the issues is that we were trying to keep the name of the app AND
the way that name displayed on a phone the same. We wanted the name on the
phone to start with 'Bible'... hence 'Bible Study App'. That was probably
an unnecessary requirement. YouVersion just calls their app 'Bible' on the
device. We could have done the same.

The other issue was to try to identify the app as a study tool. YouVersion
for example is not really a study Bible - it is devotional bible app.

'And Bible' always bothered me because 'And' is such a common English word
and never looked right to my eyes. What does this mean: "And Bible is
great."  Am I saying in clumsy English 'and the Bible is great' or am I
saying 'the And Bible app is great'. That ambiguity bothered me. That's why
I had pushed for AndBible with no spaces - it eliminates the ambiguity.

The name 'Bible Study App' feels long to me now and so I find myself not
using it when communicating and reverting back to And Bible or AB. I could
use 'BSA'.

And yes, it is a little too generic I think... which is why we kept And
Bible in the full name.

tuomas2:
> What about "Bible Study App, by And Bible"

Example from guidelines would suggest rather "And Bible: Bible study app".

[...]

PR #2182, opened 2022-06-18
App name change 

tuomas2:
Implement #1857 (change name)


tuomas2:
Will merge when translators are ready

timbze:
Looks like this needs to be changed yet...
`<string name="app_name_medium">Bible Study (And Bible)</string>`

I guess maybe it's not needed anymore since "long" is not that long now

tuomas2:
> Looks like this needs to be changed yet... `<string name="app_name_medium">Bible Study (And Bible)</string>`
> 
> I guess maybe it's not needed anymore since "long" is not that long now

I was also thinking medium version is becoming obsolete and can be replaced with long version.

PR #2191, opened 2022-07-09
Disable the ability to insert a new line when adding a new label

agrogers:
Pushing enter when entering the name of a new label would create a new line. I don't think there is ever an instance when we want a new line in a label name.

The option I used automatically capitalises each word. This is the same option used when entering a label name to search for.

### DIFF
--- a/app/src/main/res/layout/bookmark_label_edit.xml
+++ b/app/src/main/res/layout/bookmark_label_edit.xml
@@ -27,6 +27,7 @@
 			android:layout_width="0dp"
 			android:layout_height="wrap_content"
 			android:hint="@string/label_name_prompt"
+			android:inputType="textPersonName"
 			android:textAppearance="?android:attr/textAppearanceMedium"
 			app:layout_constraintEnd_toEndOf="parent"
 			app:layout_constraintStart_toEndOf="@id/titleIcon"


app/src/main/res/layout/bookmark_label_edit.xml
```diff
@@ -27,6 +27,7 @@
 			android:layout_width="0dp"
 			android:layout_height="wrap_content"
 			android:hint="@string/label_name_prompt"
+			android:inputType="textPersonName"
```
tuomas2, line 30: Wouldn't maxLines property also work here?

agrogers: Didn't know there was a maxLines property. Sounds like it might. But name is actually what i wanted i think - i usually want each word to be capitalised

timbze: I think textPersonName disables spell check. I think maxLines would probably be the right solution... if we want to set default to capitalize all words we can also use textCapWords... but I don't really like the idea of making that default. It's harder for me to "uncapitalize" first letter if keyboard makes that default than to capitalize first letter. And many of my labels I don't like all words to be capitalized...

agrogers: Fair enough. It doesn't make much difference on the search screen. I dont mind either way.

tuomas2: Hmm. maxLines = 1 does not seem to prevent from entering multi-line string (it just does not expand the text input hiding other lines).

tuomas2:
Multiline does not seem to mess up anything:
image
image
image

So I think we can leave it as it is currently, unless we can figure out a better way to fix this.

PR #2192, opened 2022-07-09
Make the active window button the same color as the workspace

agrogers:
The blue outline around the active button (which i think might have been my suggestion from a long time ago) has never quote worked for me. It was too difficult for me to see and i would still find myself changing the document of a window i did not want to change.

Also, now that I am using colors for each workspace, the blue outline clashed with the main color.

This change makes the active window button (in the bar at the bottom) the same color as the workspace color. It also applies the color to the window button itself (in the top right of each window). It is much more pleasing to my eye now. 

One issue is when the workspace color is set to gray. The gray color gets used as the Active button color which means that it doesnt look active. Not sure what to do about that. Is there some easy way to identify 'gray'?

image
image
image


tuomas2:
I thought this is made on top of #2007 but it isn't?

agrogers:
It is. And I pulled that code into this PR. Wasn't sure if this idea would
be popular so kept it separate. I like it - its like candy for my eyes now.
I want to add an option in the advanced settings to enable this
functionality. I can understand some (boring people :-) may not like it.

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

tuomas2:
> It is. And I pulled that code into this PR. Wasn't sure if this idea would be popular so kept it separate. I like it - its like candy for my eyes now. I want to add an option in the advanced settings to enable this functionality. I can understand some (boring people :-) may not like it.

If I change target branch to `Modified-workspace-color` it will have 250 commits. So there's something weird going on in these branches. If this branch feature is based on that feature, target branch should be `Modified-workspace-color`, not develop.

agrogers:
Sorry Tuomas. I don't know what I need to do to remedy this. I could make a
new branch altogether and get all changes into that. Not sure that that
resolves the issue though.

tuomas2:
I have merged develop to this branch. 

It should be updated to the current code base before merging. 
Again I don't like adding option for this in settings, if we can't get it working satisfactorily without option I think we should not merge this.

agrogers:
I am easy either way.

Why don't you like Options?

tuomas2:
> I am easy either way. Why don't you like Options?

Settings screen has already so many options that it is not easy to know/remember all the tweaks that can be done and then find the correct one. I like options, but I don't like options that <0.1%  will actually use. All added options need also to be tested and maintained.

 (We could consider doing some "advanced experimental options" like Chrome flags (chrome://flags/) that is absolutely hidden from normal user, and also leave those strings untranslated... but I'm not pro that either unless there's high enough demand)

agrogers:
Yeah, they are fair points.

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

agrogers:
Is there anything that I need to do for this PR now?

tuomas2:
it does not work as is, so yes

agrogers:
This PR is replaced by #2468

PR #2193, opened 2022-07-10
Add an option to show the long name of the book in the verse selector

agrogers:
There is not a lot of space to show the fullname of the book. But it is just readable and I think it is worth having. It might be better on tablet?

Also, removed the duplicate code that was used to display books in both grouped and standard format. 

image

image

image


oxr463 (approved):
This looks great to me!

This approach didn't modify the layout at all.

agrogers:
I modified it a little so that it now sizes the long book name based on the number of characters in any word in the name. It makes them more readible but looks a little messier. 

image

tuomas2:
Looks nice (screenshots, did not look yet the code)! Thanks!

app/src/main/java/net/bible/android/view/util/buttongrid/ButtonGrid.kt
```diff
@@ -431,6 +461,7 @@ class ButtonGrid constructor(context: Context, attrs: AttributeSet? = null, defS
     companion object {
         private const val PREVIEW_HEIGHT_DIP = 70
         private const val TAG = "ButtonGrid"
+//        enum class GRID_BUTTON_TYPES {BOOK, CHAPTER, VERSE}
```
tuomas2:
```suggestion
```

app/src/main/java/net/bible/android/view/util/buttongrid/ButtonGrid.kt
```diff
@@ -241,12 +210,73 @@ class ButtonGrid constructor(context: Context, attrs: AttributeSet? = null, defS
...
+        var smallTagStart = ""
+        var smallTagEnd = ""
+        val button = Button(context)
+        button.text = if (buttonInfo.showLongBookName && buttonInfo.type == ButtonInfo.Companion.GridButtonTypes.BOOK) {
+            // Check the length of the words in the long description and set the <size> tag accordingly
+            val maxLengthForSmall = 5
+            val maxLengthForXSmall = 9
+            val words = buttonInfo.description?.split("\\s".toRegex())?.toTypedArray()
+            var smallCnt = 1
+            words?.map { if (it.length > maxLengthForXSmall) {smallCnt = 3} else if (it.length > maxLengthForSmall) { smallCnt=2}  }
```
tuomas2: Can you explain this logic here? Effectively only comparison with the length of last word would be enough as smallCnt is overwritten all the time

app/src/main/java/net/bible/android/view/util/buttongrid/ButtonGrid.kt
```diff
@@ -241,12 +210,73 @@ class ButtonGrid constructor(context: Context, attrs: AttributeSet? = null, defS
...
+            button.typeface = Typeface.DEFAULT_BOLD
+            button.setTextSize(TypedValue.COMPLEX_UNIT_SP, textSize + 1.toFloat())
+            button.paintFlags = button.paintFlags or Paint.UNDERLINE_TEXT_FLAG
+            button.isPressed = true
+        } else {
+            button.setTextSize(TypedValue.COMPLEX_UNIT_SP, textSize.toFloat())
+        }
+        // set pad to 0 prevents text being pushed off the bottom of buttons on small screens
+        // set left and right pad to 3 prevents text from appearing off the button background
+        button.setPadding(4, 0, 4, 0)
```
tuomas2: these should be converted from dips to pixels (from CommonUtils:convertToDipsToPx)

app/src/main/java/net/bible/android/view/util/buttongrid/ButtonGrid.kt
agrogers: Yeah, I fixed that but it seems like i didnt push my fix. The logic was supposed to be 'find the longest word and set the size based on it'.

agrogers: I think your change has the same problem as my original attempt, only your version takes six lines to make the mistake i made in 1 :) But my approach seems to be a clumsy way of doing it. I am sure there is a smarter way. We just want to find the length of the longest word and assign a size based on it. And I didnt mean to commit any changes below. I still dont know exactly what i am pulling and pushing sometimes.

agrogers:
Hmm, i dont think icaused this....
`net.bible.android.control.page.window.WindowControlTest > testGetMinimisedWindows FAILED
    kotlin.io.FileAlreadyExistsException: /tmp/robolectric-Method_testGetMinimisedWindows18437596037165074080/net.bible.android.activity-dataDir/databases/andBibleDatabase.db -> /tmp/robolectric-Method_testGetMinimisedWindows18437596037165074080/net.bible.android.activity-dataDir/files/database_backups/dbBackup-0-20220720-085252.db: The destination file already exists.`

app/src/main/java/net/bible/android/view/util/buttongrid/ButtonGrid.kt
tuomas2: Fixed in 91441133d14986712845820873e3350242c1de07

tuomas2:
Is this unusable with small screens or why does it need to be optional?

agrogers:
It usable on my 6" phone. Maybe some people don't like change and prefer
the old style.

If the option is OFF by default then many people will not discover it.

If the option is provided then it makes the app just a little more
complicated.

But it is so easy to add options that I think providing it covers all bases

tuomas2:
> It usable on my 6" phone. Maybe some people don't like change and prefer
> the old style.
> 
> If the option is OFF by default then many people will not discover it.
> 
> If the option is provided then it makes the app just a little more
> complicated.
> 
> But it is so easy to add options that I think providing it covers all bases

I would not make this optional. I can't see why someone wouldn't prefer your version. Do you?

agrogers:
I think it is beautiful. So, I can't think of a reason. But some people
just dont like change. I dont mind either way.

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

tuomas2:
> I think it is beautiful. So, I can't think of a reason. But some people just dont like change. I dont mind either way.

Then let's remove optionality.

tuomas2:
image

This does not work very well in (some) other languages

tuomas2:
> This does not work very well in (some) other languages

That would, of course, be a reason to make it optional (and disabled by default)

agrogers:
It works terribly in some languages like Myanmar. Actually it might be the main names that are the problem as well. It all looks terrible. It is a good reason to provide a list based selector.

tuomas2:
Wow...

tuomas2:
Is it good idea to even merge this as we are so international? Or maybe enable only for english users?

agrogers:
An option to enable it let's the user choose. I like options.

My plan was to just truncate long names to a certain number of chars. Not
sure how that would look but would be acceptable to my eye in English

PR #2227, opened 2022-09-03
Remove verse number from first verse when sharing or copying

agrogers:
The first-verse verse number is always shown when the reference is displayed at the end of the verse. If the reference is positioned at the front of the verse, the first-verse verse number is hidden.


tuomas2:
I think this can be put into current-stable too

tuomas2:
Ah ok it looks like this code is already modified in develop, so not changing in current-stable. Merging..

PR #2228, opened 2022-09-03
Show verse reference in window buttons

agrogers:
Sometimes I want to know the reference for other windows, especially when writing notes that refer to these verses. Amazingly I can forget that reference in a matter of seconds. This PR adds an option to show the reference in the button.

There is at least one problem with this code - I don't know how/when to refresh the buttons when the app starts. But after a window is scrolled everything looks fine.

image

image


tuomas2:
Looks nice. Again I am preliminarily against making it optional. I guess it just does not hurt to show it like that?

agrogers:
It removes the pin icon to make space for the verse reference. That's the
only backward step. The icon is not needed because the button shape and
location conveys the same info.

I personally like being able to customise things to my liking. But I would
also default the option to True so people know it is there. But I think
most people would be fine with that change.

tuomas2:
> It removes the pin icon to make space for the verse reference.

Ah then need to think about it a bit more...

agrogers:
Closed by another PR that does the same,

PR #2229, opened 2022-09-05
Resolve "Unsafe use of a nullable receiver" warnings

uniqueid001:
The Android SDK now annotates java.io.File.listFiles() as nullable.


tuomas2:
Could take to current-stable

Issue #2092, opened 2022-02-12
[Developer-documentation] Add Instructions on Building in the Terminal via Docker


oxr463:
```
# Fetch source
git clone https://github.com/AndBible/and-bible.git
cd and-bible

# Run Android SDK container
docker run --rm --user root -v $(pwd):/usr/src/and-bible -it runmymind/docker-android-sdk:alpine-standalone /bin/sh

# Add extra dependencies required for development
apk add git npm

# Set permissions to the android user:group
chown -R android:android /usr/src/and-bible

# Switch to android user
su - android

# Navigate to the source directory
cd /usr/src/and-bible

# Set environment variable
export ANDROID_SDK_ROOT="/opt/android-sdk-linux"

# Compile a debug image
./gradlew assembleDebug
```

Reference(s):
- https://github.com/AndBible/and-bible/wiki/Developer-documentation
- https://github.com/mindrunner/docker-android-sdk

Originally posted as a comment, (See: https://github.com/AndBible/and-bible/pull/2091#issuecomment-1037382934).

tuomas2:
We can add Dockerifle to the project root too in case someone wants to build inside docker.

oxr463:
I can submit a quick PR for that.

PR #2230, opened 2022-09-05
Create a Docker image for AndBible development

uniqueid001:


timbze:
Is this basically the same idea as #2092?

uniqueid001:
> Is this basically the same idea as #2092?

Yeah, pretty much.  (I wasn't aware of that)  That one builds from the command-line only, while this installs and is able to run Android Studio.

docker/README
```diff
@@ -0,0 +1,37 @@
+About:
+
+AndBible's build system requires a newer version of nodejs than is packaged
+for Ubuntu 20.04.  As a result I can no longer build AndBible.  I created
+this Docker image based on Ubuntu 22.04 so that I can use it to build
+AndBible.
```
tuomas2, line 6: Just a tip: you can install any node versions on user level with nvm https://github.com/nvm-sh/nvm

tuomas2 (approved):

docker/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+          curl \
+          expect \
+          figlet \
+          file \
+          git \
+          less \
+          locales \
+          npm \
+          openjdk-18-jdk \
+          openssh-server \
```
tuomas2, line 21: why installing openssh-server?

tuomas2: ah, for X11 forwarding via ssh, okay...

PR #2232, opened 2022-09-06
Resolve deprecation warnings regarding StatFS

uniqueid001:
In API 18 the StatFS methods getBlockSize() and getAvailableBlocks()
were deprecated in favor of new methods returning `long` instead of `int`.
Additionally, API 18 added the method getAvailableBytes() which performs
the multiplcation of the block size and available blocks.  Since
AndBible's minSdk is currently 21, that method can be used without
checking the platform's version.


tuomas2:
could merge to current-stable

PR #2282, opened 2022-10-10
Start the calculator activity first, when applicable

uniqueid001:
When applicable, start the calculator activity before showFirstLayout().

### DIFF
--- a/app/src/main/java/net/bible/android/view/activity/StartupActivity.kt
+++ b/app/src/main/java/net/bible/android/view/activity/StartupActivity.kt
@@ -227,6 +227,10 @@ open class StartupActivity : CustomTitlebarActivityBase() {
 
         initializeDatabase()
 
+        // For the 'discrete' variant, go directly to the calculator first,
+        // even when there are no Bible documents already installed.
+        gotoCalculator()
+
         if (swordDocumentFacade.bibles.isEmpty()) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/net/bible/android/view/activity/StartupActivity.kt
```diff
@@ -227,6 +227,10 @@ open class StartupActivity : CustomTitlebarActivityBase() {
 
         initializeDatabase()
 
+        // For the 'discrete' variant, go directly to the calculator first,
+        // even when there are no Bible documents already installed.
+        gotoCalculator()
```
tuomas2, line 232: I don't think this is good because user need to have app set up correctly and also must manually enable calculator feature & set pin code to utilize it. 

What we could do is to let user switch on calculator feature already in startupactivity.

uniqueid001: I updated my comment, since it was not entirely accurate as-is.

This change doesn't affect the standard variants: when the "show_calculator" setting is false (or unset), then gotoCalculator() still returns without doing anything.  The difference is only when "show_calculator" is set to true, but there happens to be no bible documents.  
This is needed, along with #2286, for the discrete variant to be discrete every time it is run.

PR #2283, opened 2022-10-10
Centralize all of the tests of the BuildConfig's FLAVOR

uniqueid001:
Do not repeat the string operations to identify the build's flavor, and centralize all of the string operations in a single file.


tuomas2:
looks like a nice code improvement, merged! Thanks.

PR #2286, opened 2022-10-10
Automatically start calculator for discrete flavor

uniqueid001:
(note: this PR should consist of only one small commit (2b2e963) but with a dependency on #2285, to make the review easier)

### DIFF
--- a/app/src/main/java/net/bible/android/view/activity/StartupActivity.kt
+++ b/app/src/main/java/net/bible/android/view/activity/StartupActivity.kt
@@ -60,6 +60,7 @@ import net.bible.android.view.activity.download.FirstDownload
 import net.bible.android.view.activity.installzip.InstallZip
 import net.bible.android.view.activity.page.MainBibleActivity
 import net.bible.android.view.util.Hourglass
+import net.bible.service.common.BuildVariant
 import net.bible.service.common.CommonUtils
 import net.bible.service.common.CommonUtils.checkPoorTranslations
 import net.bible.service.common.CommonUtils.json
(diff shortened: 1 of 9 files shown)


app/src/main/java/net/bible/android/view/activity/StartupActivity.kt
```diff
@@ -312,7 +313,7 @@ open class StartupActivity : CustomTitlebarActivityBase() {
     }
 
     private suspend fun gotoCalculator() {
-        val show = CommonUtils.settings.getBoolean("show_calculator", false)
+        val show = BuildVariant.Appearance.isDiscrete || CommonUtils.settings.getBoolean("show_calculator", false)
```
tuomas2, line 316: This does not make sense to me because calculator need to be set up first (pin code etc) and is completely optional feature for discrete user too. Shouldn't it be? User must understand what he/she is doing before using this.

uniqueid001: I would say that the calculator is not optional for the discrete variant, and that choosing to install the discrete variant APK is the user understanding that they do want the app to hide itself on startup.
Certainly the discrete variant is at odds with the UX goal of discoverability, and certainly the users need a way to learn what they can do with the app.  I would suggest that word-of-mouth may be one of the best ways for this to be shared, and it can be documented where it is distributed (although we don't want that to become too well-known to those who are adversarial).

agrogers:
I thought the whole point of the discrete variant was that it forced the
use of the calculator.

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

tuomas2:
> I thought the whole point of the discrete variant was that it forced the use of the calculator.

Main point is that app name and app id can't be changed dynamically. But I see a point in making calculator forced too. (though perhaps we can allow user to disable calculator by setting pin empty... some might use discrete version for some reason in an environment in which they don't need it).

PR #2289, opened 2022-10-11
Install 'patch' in the Docker image

uniqueid001:


tuomas2:
why is this needed? merging as I believe you have a reason but please always explain this kind of non-obvious changes :-)

uniqueid001:
Sorry, I didn't realize it was non-obvious.
patch is a program for apply a "patch" (or "diff", it's the same format).  I will often use it to apply parts of changes that I have stashed in git.  'git apply' is similar, but not exactly the same.

tuomas2:
> Sorry, I didn't realize it was non-obvious. patch is a program for apply a "patch" (or "diff", it's the same format). I will often use it to apply parts of changes that I have stashed in git. 'git apply' is similar, but not exactly the same.

I do know what patch is but not why it was added here.

PR #2328, opened 2022-10-25
Memory leak: notify the BibleGestureListener to unregister from the E…

uniqueid001:
Run AndBible, then rotate the device a few times.  Using the profiler, observe that several instances of MainBibleActivity have leaked, because the associated BibleGestureListener object registered with the ABEventBus but never unregistered from it.


tuomas2:
Good work profiling these leaks!

Issue #2316, opened 2022-10-22
Discrete mode should not open links without asking user


tuomas2:

uniqueid001:
There are also links in the Help dialog, and the licenses dialog, and possibly in other places too.

tuomas2:
> There are also links in the Help dialog, and the licenses dialog, and possibly in other places too.

Hmm. If they are "links" they are more difficult to fix.

PR #2329, opened 2022-10-25
Launch in the activity's lifecycleScope, not GlobalScope

uniqueid001:
This will ensure that the job is canceled when the activity is destroyed.  Jobs that are launched in the GlobalScope are leaked if they do not complete and are not otherwise explicitly canceled. This job is leaked in the GlobalScope when the device is rotated while the confirmation dialog is displayed, because it neither completes nor is canceled.  When launched in the activity's lifecycleScope, it is canceled when the activity is destroyed, due to device rotation, while the confirmation dialog is displayed.
(#2316)


uniqueid001:
Related resource: ["The reason to avoid GlobalScope"](https://elizarov.medium.com/the-reason-to-avoid-globalscope-835337445abc), by Roman Elizarov

PR #2334, opened 2022-10-27
Use lifecycleScope instead of GlobalScope (almost) everywhere

tuomas2:


tuomas2:
I'm merging this, but could you @uniqueid001 check and see if I was doing the right thing

uniqueid001:
I'll take a look.  I had in mind to work my way through each use of GlobalScope, but you beat me to it!

PR #2344, opened 2022-11-06
Compact bookmark list view

### LINKED ISSUE #2343: Timestamp in Bookmark list view gets too little space
**Describe the bug**

Under certain conditions, probably when both the reference and the
timestamp are too long and with a large-enough font, the bookmark list
view splits the space unevenly - the reference gets too much space and
the timestamp therefore needs many lines.

4.0.651 (from fdroid)

**To Reproduce**
Not sure. When I tried fixing it, Tried with the emulator included with Android Studio, and failed to reproduce it with all of the 3 "machines" I tried (Pixel 2, Pixel 5, Nexus 4).

My own machine is a Samsung A31, fonts configured via Settings to be the largest available there. All with the base OS stuff - no custom fonts/settings/etc.

Anyway, to reproduce:
1. Configure your phone to use a large font.
2. Create a bookmark to a long ref.
3. Perhaps this happens more/only in Hebrew. Not sure.
4. Go to the bookmarks list.

I am including two screenshots - one showing the problem and another, on the same ref, showing the behavior with my proposed fix.

I am not an android developer, and this is my first patch for an Android project. Someone more experienced might want to refine my fix before merging it. It does work for me, though, on both my own phone and the emulator (where it's merely harmless - failed to reproduce there, as noted above).

didib:
Under certain conditions, probably when both the reference and the timestamp are too long and with a large-enough font, the bookmark list view splits the space unevenly - the reference gets too much space and the timestamp therefore needs many lines. Fix this by splitting the space more evenly.

Fix #2343 .

### DIFF
--- a/app/src/main/res/layout/bookmark_list_item.xml
+++ b/app/src/main/res/layout/bookmark_list_item.xml
@@ -50,11 +50,13 @@
             <TextView android:id="@+id/verseText"
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
+                android:layout_weight="1"
                 android:textAppearance="?attr/textAppearanceListItem"
                 />
             <TextView android:id="@+id/dateText"
-                android:layout_width="fill_parent"
+                android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
+                android:layout_weight="1"
                 android:gravity="end"
                 />
         </LinearLayout>


agrogers:
Thanks for contributing

tuomas2:
Thank you! I made similar change based on current-stable branch, so this is essentially merged now.

Issue #39, opened 2015-08-24
Add a way to add custom document repositories


nikaro:
Hello. It should be nice to have the ability to add other repos than the hardcoded ones.

Updates from @tuomas2:
 - Show SWORD MRL (master repository list) repositories automatically.

g4jc:
I second this. I actually keep a locally cached copy to avoid re-downloading the same books over and over, however I'm unable to add it to the hardcoded file without re-compiling the program.
https://github.com/mjdenham/and-bible/blob/595e5bfe671b482ec0a1d8f16d6fbe52229b730c/jsword-tweaks/src/main/resources/InstallManager.plugin

g4jc:
I found one work around, per: https://groups.google.com/forum/?_escaped_fragment_=topic/and-bible/hcrgU_Eqkuk#!topic/and-bible/hcrgU_Eqkuk

1) Mirror your favorite repo: wget -r --no-parent --reject "index.html*" http://path.to.some.repos/
(e.g. ftp.xiphos.org/pub/sword/)

Then on the Android's sdcard -  usingTerminal Emulator or Ghost Commander: 
cp modules and mods.d directories to /storage/sdcard0/jsword

When you boot AndBible the offline books work! :+1:

nikaro:
Thanks for the tips :+1:

mjdenham:
Another way would be to store the zips somewhere on your network and use And Bible's new 'Install from Zip' feature.

Btw which repos and modules do you use that aren't already available.  My aim is to make everything that is most useful readily available in the simplest way.

g4jc:
My biggest reason for wanting the feature was to load the Xiphos bible and dictionary modules from WLAN onto several different android devices. Since I have a slow internet it would save my a considerable amount of time. I'm not that sure those modules are in zip format? 

Thanks.

ildar:
Guys, I don't feel this feature request to be finished. There should be a
way to add custom repos. I agree that typing isn't the most fun way to do
it. Then some container format could be used.
Reopen?

johndudeck:
Greetings. I am just getting involved with module development for AndBible.

I am involved with a project to create a set of French Bible study materials for French-speaking African pastors. There are potentially more than a hundred modules that may make up this library. Many of them are not public domain and are used by permission of the publishers for distribution in this project.

[...]

PR #2427, opened 2022-12-27
Replace hard-coded URLs of SWORD repositories

uniqueid001:
Compute the (base) URLs for the configured SWORD repositories to display in the Toast message to the user.

The only effect that users should notice, from this change, is that the URLs are now sorted.  This relates to #39.  I have some subsequent changes that I am preparing to submit.


tuomas2 (approved):
Looks good!

PR #2428, opened 2022-12-27
Refactor repositories

uniqueid001:
I moved duplicate code from the subclasses to the base class, added constructor parameters, and then converted the subclasses to factory functions.  This simplifies and shortens the code.  Related to #39.

### DIFF
--- a/app/src/main/java/net/bible/service/download/Repositories.kt
+++ b/app/src/main/java/net/bible/service/download/Repositories.kt
@@ -26,12 +26,17 @@ import org.crosswire.jsword.book.sword.SwordBookMetaData
 
 abstract class RepoBase(
     val repoName: String,
+    private val supportedDocumentsFilter: BookFilter,
     ) {
     lateinit var repoFactory: RepoFactory
 
     private val downloadManager get() = repoFactory.downloadManager
 
-    abstract fun getRepoBooks(refresh: Boolean): List<Book>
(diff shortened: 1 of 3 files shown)


app/src/main/java/net/bible/service/download/Repositories.kt
```diff
@@ -53,83 +59,23 @@ abstract class RepoBase {
...
-    }
-
-    override val repoName: String get() = REPOSITORY
-
-    companion object {
-        private const val REPOSITORY = "AndBible Extra"
-        private val SUPPORTED_DOCUMENTS: BookFilter = AcceptableBookTypeFilter()
-    }
-}
+fun AndBibleRepo() = Repository("AndBible", AcceptableBookTypeFilter())
```
tuomas2, line 62: Looks good! I'm not sure if we need to keep these as factory functions, or if one instance is enough. Merging and checking if it can be changed...

tuomas2:
@uniqueid001 Please check my improvements in the above mentioned commit!

uniqueid001:
I wouldn't prefer to create global variables.  We could, however, inline those calls into RepoFactory, which is the only place where they are used.  This would still allow the memory to be garbage collected when it is not needed, RepoFactory would still be the factory that creates the repository objects, and it can also be responsible for loading user-defined repositories from the database when that is implemented.  (PR #2438)

Issue #238, opened 2019-01-24
Download Documents improvements


tuomas2:
- Search by document name / document id / document abbreviation / language

tuomas2:
Language selector could be somewhat similarly as in Android system language selection. That would make it searchable too.

tuomas2:
Started working on this and build 356 contains first batch of changes. 

Now we should collect list of recommended documents. Please let me know what do you think should be included in this list?

tuomas2:
Current view: 
![Screenshot_1585848619](https://user-images.githubusercontent.com/5811789/78279854-d4ee1f00-7520-11ea-8083-d400e1c9e490.png)

tuomas2:
The rest of items I will move to 3.4 milestone.

PR #2429, opened 2022-12-27
Replace the Refresh MenuItem in DownloadActivity with a SwipeRefreshLayout

uniqueid001:
#238: replace the refresh MenuItem with a SwipeRefreshLayout

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -291,6 +291,7 @@ dependencies {
     implementation("androidx.preference:preference:1.2.0")
     implementation("androidx.preference:preference-ktx:1.2.0")
     implementation("androidx.recyclerview:recyclerview:1.2.1")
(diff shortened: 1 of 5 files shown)


tuomas2:
I think we have reached a point when I would like to take only bugfixes to current-stable and focus everything else to develop. So changing target.

tuomas2:
Did not try yet but SwipeRefreshLayout there sounds like a good idea!

tuomas2 (approved):

tuomas2:
@uniqueid001 please check small code style improvements in the above mentioned commit

uniqueid001:
> @uniqueid001 please check small code style improvements in the above mentioned commit

I see them.  
Many years ago I learned, from another project I worked with, to align the blocks (similar to Python).  Of course Kotlin's trailing lambda grammar doesn't allow aligning the opening brace, but I still habitually maintain the alignment of the closing brace.

PR #2438, opened 2022-12-31
Move the repository global variables into RepoFactory

uniqueid001:
Allow RepoFactory to retain responsibility for creating the repository objects, and allow the objects to be garbage collected when they are not in use.

Also,  reduce coupling by passing the DownloadManager to the Repository constructor.  This eliminates the dependency from Repository to RepoFactory, which was used solely to obtain a reference to the DownloadManager.


tuomas2 (approved):
Looks good! Merging.

PR #2440, opened 2023-01-03
Use launch{} for parallelization of functions returning Unit

uniqueid001:
The `coroutineScope{}` builder waits until all of its child jobs are complete before returning.  Therefore, using `launch{}` instead of `awaitAll(async{})` avoids creating unnecessary `Deferred<Unit>` objects while achieving the same concurrency.

(documentation reference: https://kotlinlang.org/docs/coroutines-basics.html#scope-builder)

### DIFF
--- a/app/src/main/java/net/bible/android/view/activity/download/DownloadActivity.kt
+++ b/app/src/main/java/net/bible/android/view/activity/download/DownloadActivity.kt
@@ -109,6 +109,15 @@ open class DownloadActivity : DocumentSelectionBase(
     private val booksNotFound = ArrayList<String>()
     private val docDao get() = DatabaseContainer.db.swordDocumentInfoDao()
 
+    /**
+     * Load each of the document lists in parallel.
+     */
+    private suspend fun downloadDocJson() = coroutineScope {
+        launch { loadRecommendedDocuments() }
+        launch { loadDefaultDocuments() }
+        launch { loadPseudoBooks() }
+    }
+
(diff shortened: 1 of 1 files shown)


tuomas2:
> The `coroutineScope{}` builder waits until all of its child jobs are complete before returning. Therefore, using `launch{}` instead of `awaitAll(async{})` avoids creating unnecessary `Deferred<Unit>` objects while achieving the same concurrency.

Hmm this is new to me. Need to re-check. Any other benefit than code improvement?

uniqueid001:
> > The `coroutineScope{}` builder waits until all of its child jobs are complete before returning. Therefore, using `launch{}` instead of `awaitAll(async{})` avoids creating unnecessary `Deferred<Unit>` objects while achieving the same concurrency.
> 
> Hmm this is new to me. Need to re-check. Any other benefit than code improvement?

There is the (small) benefit of not creating the `Deferred<Unit>` objects, but mainly just code improvement.
I just added a link to the documentation, which I forgot to include when I submitted this last night.

timbze:
I didn't know that `coroutineScope{}` waits for all child jobs to complete. Looks like a good improvement to me (if the jobs are not required to run consecutively)

tuomas2:
I once learned from ["The Zen of Python"](https://peps.python.org/pep-0020/) that "Explicit is better than implicit". I'm thinking the code is changing from explicit to implicit if we apply this change. So in that sence it is not a code improvement. Though it is good to understand how coroutineScope works (which I did not) so I'm happy to learn that.

uniqueid001:
Okay

Issue #2402, opened 2022-12-03
Show the search button in portrait mode on a phone


agrogers:
I use the search option a lot but to get to it I need to go through the hamburger menu. I would prefer to use a button at the top of the screen. There is not much room there unfortunately. I think I would prefer the search button over the Original Language button.

Request raised here:
https://support.andbible.org/scp/tickets.php?id=1692

timbze:
#195 would probably solve this nicely. Otherwise I don't have a feeling that default buttons will be changed much, unless there's clearly improvement and not bad for existing users.

agrogers:
I think one way to do it would be to allow the user to choose an order for
the buttons. Or perhaps add another gesture to open the search.

tuomas2:
User was satisfied when he got to know that he can access search through main menu. Also I think it is sufficient.

PR #2444, opened 2023-01-04
Tap with 2 fingers to show the search dialog

agrogers:
I have always thought that the search button is too far away from the my normal workflow. Going into the hamburger menu just felt wrong and slow (that does go to show how lazy I have become!).  A two finger tap is unused (as far as I am aware) and is a nice shortcut for this. I think we could also have 3 finger taps but that will require a fancier implementation that I have done. Maybe there is justification to allow the user to customise what a two finger tap does. 

This partially addresses #2402

### DIFF
--- a/app/src/main/java/net/bible/android/view/activity/page/BibleView.kt
+++ b/app/src/main/java/net/bible/android/view/activity/page/BibleView.kt
@@ -25,6 +25,7 @@ import android.content.pm.ResolveInfo
 import android.graphics.Rect
 import android.net.Uri
 import android.os.Build
+import android.os.Bundle
 import android.os.Looper
 import android.text.TextUtils
 import android.util.LayoutDirection
@@ -46,12 +47,15 @@ import android.webkit.WebView
 import android.webkit.WebViewClient
 import androidx.annotation.RequiresApi
 import androidx.core.view.GestureDetectorCompat
+import androidx.lifecycle.findViewTreeLifecycleOwner
+import androidx.lifecycle.lifecycleScope
 import androidx.webkit.WebViewAssetLoader
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import kotlinx.serialization.Serializable
+import kotlinx.serialization.SerializationException
 import kotlinx.serialization.Transient
 import kotlinx.serialization.serializer
 import net.bible.android.activity.R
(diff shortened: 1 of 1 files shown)


tuomas2:
Common users will never ever figure this out... but I'll give it a try.

app/src/main/java/net/bible/android/view/activity/page/BibleView.kt
```diff
@@ -686,7 +685,7 @@ class BibleView(val mainBibleActivity: MainBibleActivity,
             "?lang=$lang&fontModuleNames=$fontModuleNames&styleModuleNames=$styleModuleNames&featureModuleNames=$featureModuleNames&rtl=$isRtl&night=$nightMode")
     }
 
-     fun onEvent(e: ReloadAddonsEvent) {
+     fun onEvent() {
```
tuomas2: I think this change is unintentional? Breaks reloading addons...

tuomas2:
```suggestion
     fun onEvent(e: ReloadAddonsEvent) {
```

tuomas2:
To avoid merge conflict and make reviewing easier, please do not use auto-reformat in feature / bug fix branches in general. It's ok to do reformatting separate PR's though.

app/src/main/java/net/bible/android/view/activity/page/BibleView.kt
agrogers: > I think this change is unintentional? Breaks reloading addons...
Yeah, good find. Thanks.

agrogers:
> Common users will never ever figure this out... but I'll give it a try.

It's as findable as the swipe left, right and down on the Verse Selector I think. They all need introducing. I would love to include one of those guided tutorials for new users and when new features are added. Or even just a text message...

tuomas2:
I'm having many accidental two finger taps with my tablet that I use with two hands...

agrogers:
I haven't had any issues with it yet. But I only use my phone. I liked it
when I have needed. I wouldn't see any need for it on a tablet because you
have the search button (I think) in the action bar. I think that is the
best solution to my problem, but there is no room left for a search button
on my phone.

timbze:
I can see the handiness of this... but it also feels like an unnatural shortcut... Not that I have a problem with it necessarily though. This feels like a feature that should be enabled by an advanced setting that regular users never see.

tuomas2:
I agree with Timmy, this is not good to be enabled by default.

tuomas2:
Created issue for this #2464.

tuomas2:
I already had forgotten about this and have been recently wondering why I'm always going to search unintentionally.  Weird behavior. I think I must revert this (at least until #2464 is done).

PR #2468, opened 2023-01-21
Set the active window button to the same colour as the toolbar

agrogers:
This replaces the previous PR which I will try to delete now.

### DIFF
--- a/app/src/main/java/net/bible/android/view/util/widget/WindowButtonWidget.kt
+++ b/app/src/main/java/net/bible/android/view/util/widget/WindowButtonWidget.kt
@@ -19,6 +19,7 @@ package net.bible.android.view.util.widget
 
 import android.annotation.SuppressLint
 import android.content.Context
+import android.graphics.drawable.Drawable
 import android.util.AttributeSet
 import android.view.KeyEvent
 import android.view.LayoutInflater
@@ -32,6 +33,7 @@ import net.bible.android.control.page.window.Window
 import net.bible.android.control.page.window.WindowChangedEvent
 import net.bible.android.control.page.window.WindowControl
 import net.bible.android.view.activity.download.imageResource
+import net.bible.android.view.activity.page.MainBibleActivity
 import net.bible.service.common.CommonUtils
 import net.bible.service.common.CommonUtils.getResourceColor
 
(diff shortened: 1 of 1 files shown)


tuomas2:
Allright let's try..

tuomas2:
Colors are very confusing in restore window area.
https://youtu.be/Y5a2xBLV_Go

tuomas2:
will revert now. You can create a new pull request, reverting the revert commit  (c737a5aa26838cfb2ab055b23250d243837ea57b) and then improving this further.

agrogers:
I am not sure what the problem is.  the active window takes the toolbar
color (replaces blue outline) . The dark buttons are are visible windows
(same as before). And light buttons are not visible windows (same as
before). Only change is that the blue outline is replaced with a solid
colour.

Thinking about it now just changing the outline colour would be more
aesthetically pleasing than the blue outline which looks out of place with
new colours.

Mine looks both clear and beautiful 😇

agrogers:
I am not sure what the problem is. the active window takes the toolbar color (replaces blue outline) . The dark buttons are are visible windows (same as before). And light buttons are not visible windows (same as before). Only change is that the blue outline is replaced with a solid colour. Thinking about it now just changing the outline colour would be more aesthetically pleasing than the blue outline which looks out of place with new colours. Mine looks both clear and beautiful 😇

tuomas2:
Hmm maybe I need to give it another try with more than just 2 windows

timbze:
Another option might be to just change the border color, or even make it a bit thicker.

agrogers:
Yeah, if the border matched the workspace color it would look nice. My
original reason for the change was because I have accidentally changed the
document for the wrong window soooo many times. That always bugs me. Making
the active window button more identifiable was my aim and this helps but I
still sometimes change the document for the wrong window.

My suggestion to change the way the toolbar buttons operate will mostly
solve that problem for me.

agrogers:
@tuomas2, is this merged into the develop branch? It looks like it is but I can't find it in develop.

tuomas2:
> @tuomas2, is this merged into the develop branch? It looks like it is but I can't find it in develop.

I reverted it as I said here:
> will revert now. You can create a new pull request, reverting the revert commit (https://github.com/AndBible/and-bible/commit/c737a5aa26838cfb2ab055b23250d243837ea57b) and then improving this further.

But I will now revert my revert so it will get another try.

tuomas2:
with default grey workspace color there is no active window indication now.

agrogers:
OK, i read 'give it another try' differently. 

I will have a look at the grey worksapce color.

tuomas2:
What about reverting to old version but making border significantly thicker:

image

It match more with the style of active window highlight (blue corner markers)

tuomas2:
Implemented this in 78a9828996d9251a0d1187b5774b7746c702279f (but it's still open for discussion / different solutions)

PR #2471, opened 2023-01-21
Show the current verse reference in the window buttons

agrogers:
This replaces #2228

Sometimes I want to know the reference for other windows, especially when writing notes that refer to these verses. Amazingly I can forget that reference in a matter of seconds. This PR adds an option to show the reference in the button.

image


tuomas2:
It's getting a bit crowded with all the info, but I agree this might be handy.

tuomas2:
Let's see if one day we need to prioritize space usage with some other indicators :-P

tuomas2:
Oops, forgot to review. Will create another PR for this and let's finish it there.

agrogers:
I think it should be an Option, not mandatory.

PR #2484, opened 2023-01-28
Adjusted padding to better align number and simplified menu option slightly

agrogers:
That feels better to me.

I like the sync group option. It doesnt really change existing functionality unless someone goes hunting for the option. Limiting to 4 groups is good too. I imagine this is most useful on tablet.

The only other change I might suggest is having a new window open in the same sync group. At present it always opens in Sync Group 1. It feels like it should keep it's number. Having not used it yet I don't know if that is a good idea though.


tuomas2:
> The only other change I might suggest is having a new window open in the same sync group. At present it always opens in Sync Group 1. It feels like it should keep it's number. Having not used it yet I don't know if that is a good idea though.

This was a bug that was fixed in ac0a07f6082b5404fbf1c92d2d851928363b69a0 (sync group was not saved at all)

agrogers:
Great. I like it. It's kind of the perfect type of new
functionality. Existing users who dont want it will barely notice it. But
some will notice the little number and go exploring to see why it's there.

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

tuomas2:
I actually aligned number intentionally so that it looks a little bit like a subscript to the sync symbol. What do you think?

agrogers:
I didn't realise you had asked a question here. And I can't remember what I changed and how it looked. So you can cancel this PR.

PR #2495, opened 2023-02-02
#2471 again (show-ref-in-button)

tuomas2:
#2471 was merged too quickly...


tuomas2:
@agrogers 

Hmm I didn't notice that this replaced pin mode icon

tuomas2:
Leaving this now in broken state. 

Need to solve some issues with syncing still when restoring a window that is synced.

agrogers:
We did discuss that the pin has to go last time. The pin is unnecessary.
The button shapes themselves identify that it is pinned. And they are
grouped together.

PR #2536, opened 2023-03-04
Don't show the window buttons when scrolling

agrogers:
One thing that has bugged me for a while is that everytime I scroll each window button becomes visible. I dont think there is ever a good reason to show those buttons when scrolling. When they do appear they obscure the text that I am scrolling through and trying to read.

This change simply disables this feature. The buttons still show on click events and when a window becomes active which is exactly what I want.


tuomas2:
I'll try if this and see if it causes undesired side effects.

PR #2549, opened 2023-03-20
Allow searching in bookmarks

JJK96:
Adds a search field to the bookmarks view to filter bookmarks/notes based on the query.

**Screenshots**
image

### DIFF
--- a/app/src/main/java/net/bible/android/control/bookmark/BookmarkControl.kt
+++ b/app/src/main/java/net/bible/android/control/bookmark/BookmarkControl.kt
@@ -159,11 +159,29 @@ open class BookmarkControl @Inject constructor(
 
     fun deleteBookmarksById(bookmarkIds: List<Long>) = deleteBookmarks(dao.bookmarksByIds(bookmarkIds))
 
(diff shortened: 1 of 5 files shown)


tuomas2:
My initial feeling is that this might impact performance as we are doing notes filtering always. I would suggest that we do it only if we are searching something. I.e. separate functions for listing all bookmarks and for searching with notes.

Another performance improvement that could be good idea would be to use `@Fts4` annotation (see how `DocumentSearch` (`DocumentSearchDao.search`) is using that) to make full text search index for notes. 

How does this sound?

JJK96:
Indices are not allowed in a FTS table:
```
/Users/nw/git/and-bible/db/build/tmp/kapt3/stubs/debug/net/bible/android/database/bookmarks/BookmarkEntities.java:13: error: Indices not allowed in FTS Entity.
    public static final class Bookmark implements net.bible.android.database.bookmarks.VerseRangeUser {
```

So, I'm leaving the bookmarks table for what it is for now, because it uses some other indices currently.

JJK96:
@tuomas2 I implemented it in separate functions/queries now

tuomas2:
> @tuomas2 I implemented it in separate functions/queries now

Okay, thanks I'll check out when I have time :-)

tuomas2:
Merged develop to your branch and pushed it to our repo as I don't have permission to your fork.
As you are dev team member you can push directly to our repo (except protected branches). Closing this and continuing in #2685

PR #2551, opened 2023-03-22
Fix problematic coroutine of startPlan

### LINKED ISSUE #2141: Set as current day doesn't change start date
**Describe the bug**
It worked in some cases but did not work with the plan at location attached as screenshot below. Feature that start date would change was added in #2125

Build 643

**Screenshots**

Adriankhl:
I think this fixes https://github.com/AndBible/and-bible/issues/2141

Since `startPlan()` is executed in parallel, if the process is too slow
https://github.com/AndBible/and-bible/blob/051a4751d7e444ce4ef3da04d1ac483a8e03df68/app/src/main/java/net/bible/android/view/activity/readingplan/DailyReading.kt#L342
may completes after
https://github.com/AndBible/and-bible/blob/051a4751d7e444ce4ef3da04d1ac483a8e03df68/app/src/main/java/net/bible/android/view/activity/readingplan/DailyReading.kt#L345
which causes the failure of "Set as current day".

Actually I don't understand why there is a need to have coroutine in `ReadingPlanRepository`, at least how it is implemented seems to be quite error-prone.

### DIFF
--- a/app/src/main/java/net/bible/service/db/readingplan/ReadingPlanRepository.kt
+++ b/app/src/main/java/net/bible/service/db/readingplan/ReadingPlanRepository.kt
@@ -63,7 +63,7 @@ class ReadingPlanRepository @Inject constructor() {
     }
 
     @Synchronized
-    fun startPlan(planCode: String, date: Date = CommonUtils.truncatedDate) = scope.launch {
+    fun startPlan(planCode: String, date: Date = CommonUtils.truncatedDate) = runBlocking {
         var readPlan = readingPlanDao.getPlan(planCode)
         readPlan = readPlan?.apply { planStartDate = date } ?: ReadingPlan(planCode, date)
 


timbze:
Thank you @Adriankhl, I think you are right. I wouldn't be surprised if there were more issues like it (I was too new to coroutines when I wrote that 😦)

Did you test that it's working with this change?

Adriankhl:
At least it seems to work on my phone (Pixel 4a 5G), I never experience the issue after the change, but you know it is quite tricky to reproduce this kind of parallel process issues.

app/src/main/java/net/bible/service/db/readingplan/ReadingPlanRepository.kt
```diff
@@ -63,7 +63,7 @@ class ReadingPlanRepository @Inject constructor() {
     }
 
     @Synchronized
-    fun startPlan(planCode: String, date: Date = CommonUtils.truncatedDate) = scope.launch {
+    fun startPlan(planCode: String, date: Date = CommonUtils.truncatedDate) = runBlocking {
```
tuomas2, line 66: What does runBlocking here do differently than if we just have it as a ordinary function like

```suggestion
    fun startPlan(planCode: String, date: Date = CommonUtils.truncatedDate) {
```

?

Adriankhl: It is because it calls the functions `getPlan` and `updatePlan` which are suspending functions so we need the runBlocking block somewhere.

Adriankhl: And I just follow the style of the other functions in `ReadingPlanRepository`

tuomas2: Ah correct. Thanks for explanation.

tuomas2:
This can go also to current-stable. Cherry-picking there...

Issue #64, opened 2016-11-05
Intent filter to perform bible search / show bible links in AB


jerryOkafor:
- [x] #2559 Open bible links in AB
- [ ] Perform search via intent

Can we have this as an api, i have always needed a bible api and all i could do is to modify this bible adding an intent filter for a bible ref, maybe i will show you a little code.

mjdenham:
Yes, code in a merge request with a suitable unit test would be great.

po10cio:
After a long time,thanks. I have spent time trying to understand the architecture of the app, it a little complex, any thing to help me get on-board?

mjdenham:
Sorry not much, it is complex, it started off around the time of Android 1.6 as a proof of concept to see if I could use JSword on Android, and just grew.

I suppose that once you have added the uri intent handling it is just a matter of returning the bible text.

There are some standard Sword uris.  You will find reference to these in UriAnalyzer and use in LinkControl.  They are specified in https://www.crosswire.org/wiki/Frontends:URI_Standard

Once the link has got somewhere into the app I suppose you need to analyse it like in UriAnalyzer and LinkControl and then return the Bible text like in PageControl.copyToClipboard.

Regards
Martin

tuomas2:
- Which URL scheme to use for And Bible links? Is there generic scheme available? Or do we just use the internal scheme (e.g. sword://Bible/Gen.1.1)
- Consider offering to open bible.com links if possible. Perhaps other popular online services too.

ildar:
Wow, that's what I requested years ago in
http://code.google.com/p/and-bible/issues/detail?id=188 with "[FR] add
intent filter for
URLs: http://___some online bible site___/$ book/$ ch/$ verse/$ end verse" !

Totally agree it's a useful feature. I'm thinking of inserting such links
into some classical e-books so a reader can tap to read the passage
referenced in the text + see the context.
<http://code.google.com/p/and-bible/issues/detail?id=188>

ildar:
Regarding the Bible site URLs intercept:

1. I'd suggest to have URL intents for popular Bible reading sites
2. Is there a web version of Sword with hundreds of modules available
online?

mcepl:
There is https://github.com/Verses/Common_Bible_Share_Sheet which is for iOS, but otherwise it looks like the same thing.

agrogers:
I suggest we delay this one to after v3.4 release

mcepl:
[...]

PR #2560, opened 2023-04-06
Closes #2559 (Open bible links in AB)

### LINKED ISSUE #2559: Open bible links in AB

tuomas2:
Closes #2559 (related to #64)

When enabled, open links in AB if user goes to url's like
- https://andbible.org/bible/Gen.1.1?document=FinRK&v11n=KJV (where query parameters are optional)
- https://www.bible.com/fi/bible/1745/1SA.13.FINRK

Configuration:

image

image

### DIFF
--- a/app/src/standard/AndroidManifest.xml
+++ b/app/src/standard/AndroidManifest.xml
@@ -35,6 +35,21 @@
                 <!-- see Samsung multi-window info at bottom -->
                 <category android:name="android.intent.category.MULTIWINDOW_LAUNCHER" />
             </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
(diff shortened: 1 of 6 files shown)


app/src/standard/AndroidManifest.xml
```diff
@@ -35,6 +35,21 @@
                 <!-- see Samsung multi-window info at bottom -->
                 <category android:name="android.intent.category.MULTIWINDOW_LAUNCHER" />
             </intent-filter>
+            <intent-filter>
```
tuomas2, line 38: andbible.org could be auto-enabled if using `android:autoVerify="true"` but it would query automatically https://andbible.org/.well-known/assetlinks.json which we do not want.

tuomas2: andbible.github.io is not an option for autoverify, because github.io does not serve files starting with dot.

tuomas2:
Some documentation https://github.com/AndBible/and-bible/wiki/Bible-links

PR #2565, opened 2023-04-07
Fix StepBible link document (version)

### LINKED ISSUE #2563: StepBible link uses "version" and not "document"
This does not work:
https://www.stepbible.org/?q=reference=Gen.1.1&document=KJV&v11n=KJV

This works:
https://www.stepbible.org/?q=reference=Gen.1.1|version=KJV&v11n=KJV

Also looks like reference and version must be separated by | and not & (either reference or version can be first)

**Side note**: not sure for what v11n is in this case or if it helps.

Build 684 alpha

timbze:
Closes #2563

At least with this it works when our document name matches StepBible's name.
Our NASB is StepBible NASB1995 (they also have NASB2020). There's more differences like that.

TODO
- [x] Update wiki https://github.com/AndBible/and-bible/wiki/Bible-links

### DIFF
--- a/app/bibleview-js/src/utils.ts
+++ b/app/bibleview-js/src/utils.ts
@@ -617,5 +617,6 @@ export function sprintf(format: string, ...args: any[]) {
 
 export function formatExportLink({ref, v11n, doc}: {ref: string, v11n: string, doc?: string}) {
     const docStr = doc ? `|version=${doc}`: ""
+    // this is parsed in MainBibleActivity::openLink(uri: Uri) function for intent parsing
     return `https://stepbible.org/?q=reference=${ref}${docStr}&v11n=${v11n}`
 }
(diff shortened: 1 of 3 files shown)


app/bibleview-js/src/utils.ts
```diff
@@ -616,6 +616,6 @@ export function sprintf(format: string, ...args: any[]) {
 }
 
 export function formatExportLink({ref, v11n, doc}: {ref: string, v11n: string, doc?: string}) {
-    const docStr = doc ? `&document=${doc}`: ""
-    return `https://stepbible.org/?q=reference=${ref}&v11n=${v11n}${docStr}`
-}
\ No newline at end of file
+    const docStr = doc ? `|version=${doc}`: ""
+    return `https://stepbible.org/?q=reference=${ref}${docStr}&v11n=${v11n}`
+}
```
tuomas2: This change is not enough: we need to also intepret this formatting correctly in MainBibleActivity:openLink

tuomas2: (in case if user clicks link on a device that has andbible set to handle stepbible.org links)

timbze: I think this is fixed now, but in debug something seems problematic. It seems to start a new AB activity each time and then only opens the reference occasionally. I haven't seen this issue in release build from before though.

tuomas2:
Merging, let's test this!

PR #2566, opened 2023-04-07
Better file export with intents (study pad export and backups)

### LINKED ISSUE #1753: "Share" option for database backup might not make sense
"Storage" option also allows storing db in Google Drive etc. But Share allows sending db with email. Is it necessary? Should we just show the dialog that comes with "Storage"?

timbze:
I found a way to join file intent and share intent into the same intent.

I think this makes it much less confusing to save a backup or export study pad because now "Files" shows in the same menu with other share options.

Fixes #1753

- [x] Error report - add files option as well
- [x] Simplify messages (dialogs) in backup screen
- [x] Check ReadTheDocs documentation (https://github.com/AndBible/docs/pull/8)


app/src/main/java/net/bible/android/view/activity/page/BibleJavascriptInterface.kt
```diff
@@ -435,9 +436,30 @@ class BibleJavascriptInterface(
...
-        mainBibleActivity.startActivity(chooserIntent)
+        chooserIntent.putExtra(Intent.EXTRA_INITIAL_INTENTS, arrayOf(saveIntent))
+
+        scope.launch(Dispatchers.Main) {
+            val result = mainBibleActivity.awaitIntent(chooserIntent)
+            val data = result.resultData
+            if (result.resultCode == Activity.RESULT_OK && data?.data != null) {
+                val destinationUri = data.data!!
+
+                mainBibleActivity.contentResolver.openOutputStream(destinationUri)?.use { outputStream ->
```
tuomas2: perhaps actual saving be done in Dispatchers.IO?

tuomas2:
This should work for all supported android versions (Android 5 +)?

tuomas2:
Would this fit for error report action as well? Though primarily email is intended to be used there, not saving to a storage etc.

app/src/main/java/net/bible/android/view/activity/page/BibleJavascriptInterface.kt
timbze: Good idea. Done

timbze:
> This should work for all supported android versions (Android 5 +)?

I think so but will need to get #2568 figured out before I'll be able to test it. Maybe I can try emulator.

app/src/main/java/net/bible/android/control/backup/BackupControl.kt
```diff
@@ -390,108 +380,75 @@ object BackupControl {
...
+        val shareIntent = Intent(Intent.ACTION_SEND).apply {
+            // MIME of .apk is "application/vnd.android.package-archive".
+            // but Bluetooth does not accept this. Let's use "*/*" instead.
+            type = "*/*"
+            putExtra(Intent.EXTRA_STREAM, fileUri)
+        }
+        val saveIntent = Intent(Intent.ACTION_CREATE_DOCUMENT).apply {
+            addCategory(Intent.CATEGORY_OPENABLE)
+            type = "application/vnd.android.package-archive"
+            putExtra(Intent.EXTRA_TITLE, "and-bible.apk")
```
tuomas2, line 415: Noticing that we should have calculator.apk for calc build. Can create ticket and change myself..https://github.com/AndBible/and-bible/issues/2572

timbze:
> Maybe I can try emulator

Tried on emulator even though I had same issues, I was able to test backup and seemed to work perfectly.

app/src/main/res/values/strings.xml
```diff
@@ -659,10 +659,7 @@
     <string name="backup_email_subject_2">%s database backup</string>
     <string name="backup_email_message_2">Your %s database containing bookmarks, notes, reading plans and workspaces is attached</string>
     <string name="restore_unsuccessfull">Restoring database was unsuccessful. Perhaps you gave invalid database file.</string>
-    <string name="backup_phone_storage">Phone storage</string>
-    <string name="backup_backup_title">Backup to where?</string>
-    <string name="backup_backup_message">Backup to phone or elsewhere via Share function (email, Google Drive etc.)?</string>
-    <string name="backup_share">Share</string>
+    <string name="backup_share">Backup</string>
```
tuomas2: Need to check if this is old string in current stable version. We should not change old strings to completely different but instead create a new one with another id.

tuomas2: yep backup_share is in current stable. So maybe rename to something else?

timbze:
> Would this fit for error report action as well? Though primarily email is intended to be used there, not saving to a storage etc.

Didn't think about that. Yes I think I should implement there too. I know we have had users that did not have any of the share apps and had problem exporting log file.

app/src/main/res/values/strings.xml
timbze: Fixed now. We had backup_button already which is right context and string, so I just removed backup_share and used backup_button

timbze:
> Would this fit for error report action as well? Though primarily email is intended to be used there, not saving to a storage etc.

Implemented this now. But it just saves logcat file (not screenshot). I think saving to files can only be one file at a time, unless zipping them together. Is sharing screenshot to files important? I guess it could be helpful in a case where app has crashed and screenshot from that has been saved...

tuomas2:
screenshot still goes to email?

timbze:
> screenshot still goes to email?

Yes

timbze:
> screenshot still goes to email?

I have now updated it to zip screenshot and logcat together when saving to file. I figure that's probably better.

tuomas2:
Looks good! Thank you. Merging so we can start testing this.

Issue #2588, opened 2023-05-04
Split database into separate files


tuomas2:
- Bookmarks & Labels
- Workspaces & windows & related settings
- Reading plans
- Document search (no backup, but store in another db)
- Application preferences (settings in general)
- Book data
   - Custom repositories
   - Sword document info

Database backup file will be zip of those files. User can select which categories he/she wants to restore. Also for Google Drive syncing, it is possible to select desired categories.

PR #2589, opened 2023-05-04
Split database into multiple files

tuomas2:
Implement #2588


timbze:
Should we start to use custom file extension to more easily be able to restore through intent? With zip AB wants to open file but for modules. I guess AB could also check what type of content the zip has (if we want to keep listening to zip intents anyway for modules).

tuomas2:
> Should we start to use custom file extension to more easily be able to restore through intent?

Good point. Worth considering. 

> With zip AB wants to open file but for modules. I guess AB could also check what type of content the zip has (if we want to keep listening to zip intents anyway for modules).

I was thinking to change install zip feature such that it will be able to handle module backups also. 
It will show the content of the zip file (modules and db backups) and ask what user actually wants to install from there. So in that sense it does not matter.

timbze:
> Good point. Worth considering.

I think it might also be good for AB users to be able to easier see which file is the AB backup. We could also use the custom format for module backups... Just ideas to consider.

tuomas2:
> > Good point. Worth considering.
> 
> I think it might also be good for AB users to be able to easier see which file is the AB backup. We could also use the custom format for module backups... Just ideas to consider.

would it be confusing to allow backupping db and modules to same file?

timbze:
> would it be confusing to allow backupping db and modules to same file?

Because of huge size difference that is possible, my first thought is it would be better to always keep them separate. Maybe something like .abdb for any db backups and .abmd for modules... They could still essentially be zip format, just with different extension. If we did that for modules though we might want to have clear instructions that user knows they can unzip on desktop if they want to use in other SWORD apps. (apps like 7zip will try to unzip any file, and file could be renamed to different extension, though by default I think Windows settings are to not allow easy renaming of common extensions)

PR #2644, opened 2023-06-17
Minor beta notes gramar adjustments

meichthys:
Just a few small grammar adjustments. Hopefully there will be more substantial contributions in the future :)


tuomas2:
We are using en_GB in transifex (we do not have language dialects in AndBible so we abuse en_GB for this) for minor english grammar corrections (that do not change the meaning of sentence). This way we avoid forcing these minor changes to re-translate these strings to all languages.

I can do these changes based on this PR but after I do, I will close this pr without merging.

timbze:
Fixed in en_GB in Transifex now

PR #2685, opened 2023-07-21
Allow searching in bookmarks

tuomas2:
By @JJK96 (#2549)

Adds a search field to the bookmarks view to filter bookmarks/notes based on the query.

**Screenshots**
image


tuomas2:
So I reviewed and tested this now. Code looks more or less acceptable (can do fine tuning myself).
But functionality-wise I had conflicting thoughts about the feature.

It felt confusing that it did not try to find based on bible text nor based on bible reference. That's something that user will definately try to use this search box for. It's good that you can also search notes, but that's just one use case.

We need either to
- change ui so that it is clear to user that we are searching only based on notes content. Maybe a function in 3dot menu instead of always-visible search box?
- search also based on bible ref and bible text content

Also, if user has used any rich text formatting (html) in his notes, it won't work. Should it? Though making it work would not be trivial.

tuomas2:
Now that I'm again thinking about this, it would be best to implement separate FTS table (stored in temporary (not to be backed up) database file, same file that stores fts table for quick search for modules) for these lookups, that would contain all information needed for searching (bible ref in plain text, bible text in plain text, notes in plain text, all formatting stripped off and original bookmark id) and that would be updated based on created at and updated at timestamps in bookmark table every time bookmark activity is opened. 

Now that we are using most recent sqlite, we can use also FTS5 (https://github.com/requery/sqlite-android/). Though I haven't looked at it too much if it has benefits over fts4 that we are using and which Room library supports.

Re 

> Indices are not allowed in a FTS table:
> ```
> /Users/nw/git/and-bible/db/build/tmp/kapt3/stubs/debug/net/bible/android/database/bookmarks/BookmarkEntities.java:13: error: Indices not allowed in FTS Entity.
>     public static final class Bookmark implements net.bible.android.database.bookmarks.VerseRangeUser {
> ```
> 
> So, I'm leaving the bookmarks table for what it is for now, because it uses some other indices currently.

Yes I agree bookmarks table should not have FTS enabled. It should be separate table for searching purpose only.

JJK96:
I agree that searching for bible refs and bible text content could be useful. Also, indeed searching for text with mark-up will be tricky. So an FTS table might be useful indeed. 

However, I don't see how you would want the search in bible text content to work. Would you save the bible text of the translation that was active at the time of the note? What if you use a different bible translation later? If search in bible text content is tricky to implement I think that would not be a big problem, because we already have a bible text search in the bible view, after which also the notes of that verse can be viewed.

tuomas2:
> I agree that searching for bible refs and bible text content could be useful. Also, indeed searching for text with mark-up will be tricky. So an FTS table might be useful indeed.
> 
> However, I don't see how you would want the search in bible text content to work. Would you save the bible text of the translation that was active at the time of the note? What if you use a different bible translation later? If search in bible text content is tricky to implement I think that would not be a big problem, because we already have a bible text search in the bible view, after which also the notes of that verse can be viewed.

I'm not sure, it's just first time user experience. When I look at the list of bookmarks, I see a lot of bible text. "Search" box above indicates me that I can search the content of those bookmarks, i.e. the bible text. Doesn't it make sense? 

Bookmarks have bible version stored into them, so that should be the version that FTS table stores.

JJK96:
Updated this branch with the latest changes on develop

tuomas2:
> Updated this branch with the latest changes on develop

see commit history and changes. something is wrong.

JJK96:
Fixed commit history. The following screenshot shows that search now also works with non-bible bookmarks.

Screenshot_20230826-182805

tuomas2 (commented):
UI not super pretty but code looks ok, and functionality is an improvement.
Made couple of code improvements, you could check commit 0123619b559f50448d4c35496596c138e1ca728b. 
Changed title "Search" to "Filter by notes" so it corresponds more what it actually does.

tuomas2:
@JJK96 Thanks! See also previous comment. Merging

timbze:
> Changed title "Search" to "Filter by notes" so it corresponds more what it actually does.

I think it should be "Search notes". "Filter notes" could work too though I'd prefer the word "search" I think.

tuomas2:
Changed

JJK96:
You know I didn't implement FTS yet? I'm still working on that

tuomas2:
I just merged this PR with small changes. No FTS.

tuomas2:
FTS is still a welcome improvement if you do it :-)

Issue #2622, opened 2023-06-10
Important keyboard shortcuts


tuomas2:
Related: #1878 #1960 #60 

- [x] Go to find (ctrl+f)
- [x] Go to bible book selector (ctrl+b)
- [x] #3042
- [x] plus button in one-tap dialog
- [x] when tts panel is visible, space should toggle play/pause
- [x] workspace selector ctrl+w?
- [ ] minimize window key binding? 
- [ ] progressing ( and scrolling when necessary) with keyboard one ordinal (or verse) at time, and then enter key opening bookmark amb modal. Arrow keys would be natural but maybe that would not be good default behavior?
- [ ] when one-tap modal is open, arrow keys could move up/down in verses and shift+up/down could also select upper/lower verses?
- [ ] Make wiki page about how to use AndBible efficiently with keyboard, and link from / to Waydroid setup page
- [ ] Help dialog should list keyboard bindings and perhaps printable cheat sheet would be nice too 
- [ ] Add tooltip to buttons if there is keyboard binding to that function
- [ ] If verse is highlighted in the beginning, plus button should select one more verse and pop up single tap dialog. 
- [x] Minus button should reduce selection by one
- [ ] alt+arrow buttons move focus to other window. But when focus is moved to another window that has text edit active, you can't directly write tet to that text edit. Focus should be to text edit directly. Also it sometimes jumps to the beginning of document, which is not desired.
- [ ] esc key is harmful as it is "back key" and it is bound to going back in history. Should be  able to disable that behavior for back key (perhaps a setting) and do history navigation some other way (perhaps overlay buttons)

timbze:
Ctrl + c would be nice to have when verse is selected (modal is showing). It should copy with current copy settings i.e. the settings that are selected when pressing "Share" on the modal.

tuomas2:
bookmark amb modal:
- b: add bookmark
- n: add note
- space: start speaking

also: when pressing b when there's a selection, should make bookmark. (and note similarly).

tuomas2:
ctrl+b should make text bold in text view and is conflicting

PR #2708, opened 2023-08-14
Ctrl + c copies text with correct formatting

timbze:
Related #2622 

One issue still needs to be solved.

If I select text manually on AB startup (before verse has been clicked), and then press Ctrl + c, it uses system copy instead of my implemented "share copy". This is because `AmbiguousSelection` has not yet been loaded (which listens to Ctrl + c press).

However, once I click a verse and `AmbiguousSelection` loads, I can close the modal, select text manually and press Ctrl + c and the text is copied using my "share copy". I am not sure why, because it looks like `setupDocumentEventListener` in utils does (or should) removeEventListener on `onUnmounted`.

I would like for ctrl + c to always use my "share copy", is that correct thinking for user UX? Commentaries and genbooks etc will still just have regular system copy.

If ctrl + c should always use "share copy" even for manually selected Bible text, should the Ctrl + c listener be somewhere else than in `AmbiguousSelection` component?


tuomas2:
Are you planning to work on the remaining issue or waiting for help?

timbze:
> I would like for ctrl + c to always use my "share copy", is that correct thinking for user UX? Commentaries and genbooks etc will still just have regular system copy.
> 
> If ctrl + c should always use "share copy" even for manually selected Bible text, should the Ctrl + c listener be somewhere else than in `AmbiguousSelection` component?

I'm waiting for a response to this then I want to finish it

tuomas2:
> > I would like for ctrl + c to always use my "share copy", is that correct thinking for user UX? Commentaries and genbooks etc will still just have regular system copy.
> > If ctrl + c should always use "share copy" even for manually selected Bible text, should the Ctrl + c listener be somewhere else than in `AmbiguousSelection` component?
> 
> I'm waiting for a response to this then I want to finish it

In which context is CTRL+C supposed to work exactly? I though it is only when AmbiguousSelection component is visible.

timbze:
> In which context is CTRL+C supposed to work exactly? I though it is only when AmbiguousSelection component is visible.

I would like for it to also work when selecting text with mouse. In this case AmbigiousSelection component does not show.

tuomas2:
> > In which context is CTRL+C supposed to work exactly? I though it is only when AmbiguousSelection component is visible.
> 
> I would like for it to also work when selecting text with mouse. In this case AmbigiousSelection component does not show.

It needs separate implementation I think. It's different to work with AmbiguousSelection than general selection.

timbze:
> It needs separate implementation I think. It's different to work with AmbiguousSelection than general selection.

Ok, I know how I could do that, but I was a bit confused why the listener in AmbiguousSelection does not get unregistered when AmbigiousSelection is unloaded. Do you know why?

tuomas2:
> > It needs separate implementation I think. It's different to work with AmbiguousSelection than general selection.
> 
> Ok, I know how I could do that, but I was a bit confused why the listener in AmbiguousSelection does not get unregistered when AmbigiousSelection is unloaded. Do you know why?

AmbiguousSelection is loaded all the time, just UI part hidden most of the time (see BibleView.vue where it is loaded).

timbze:
I think this is done now

tuomas2 (approved):
Looks good, thank you!

PR #2811, opened 2023-09-05
fix choose document search #2657

### LINKED ISSUE #2657: MyBible documents do not appear in search in "Choose Document"
**To Reproduce**
1. Go to "Choose Document"
2. Look for your installed MyBible module and you should find it
3. Now search for any terms in the name or description of the MyBible module
4. -> Does not show up

Tested only on 5.0.700-beta but I believe this bug will be in production as well

RuggMatt:
MyBible documents now appear when using search on "Choose document" screen. 

Closes #2657 

**Screenshots**
Screenshot_20230904-182333

### DIFF
--- a/app/src/main/java/net/bible/android/view/activity/base/DocumentSelectionBase.kt
+++ b/app/src/main/java/net/bible/android/view/activity/base/DocumentSelectionBase.kt
@@ -427,11 +427,20 @@ abstract class DocumentSelectionBase(
                         val searchString = "${binding.freeTextSearch.text}*"
                         val osisIds = if(searchString.length < 3) null else dao.search(searchString).toSet()
 
+                        val test = { it: Book ->
+                            // matches from list of osisIds, if any
+                            (osisIds?.contains(it.osisID) ?: (searchString.length < 3)) ||
+                            // partial name match
+                            it.name.contains(binding.freeTextSearch.text,true)
+                            // partial abbreviation  match
+                            it.abbreviation.contains(binding.freeTextSearch.text, true)
+                        }
+
                         for (doc in allDocuments) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/net/bible/android/view/activity/base/DocumentSelectionBase.kt
```diff
@@ -427,11 +427,21 @@ abstract class DocumentSelectionBase(
                         val searchString = "${binding.freeTextSearch.text}*"
                         val osisIds = if(searchString.length < 3) null else dao.search(searchString).toSet()
 
+                        val test = { it: Book ->
+                            Log.d(TAG, "${it.name}")
```
RuggMatt:
```suggestion
```

app/src/main/java/net/bible/android/view/activity/base/DocumentSelectionBase.kt
```diff
@@ -427,11 +427,20 @@ abstract class DocumentSelectionBase(
                         val searchString = "${binding.freeTextSearch.text}*"
                         val osisIds = if(searchString.length < 3) null else dao.search(searchString).toSet()
 
+                        val test = { it: Book ->
+                            // matches from list of osisIds, if any
+                            (osisIds?.contains(it.osisID) ?: (searchString.length < 3)) ||
```
tuomas2, line 432: instead of osisID should we look for initials?

RuggMatt: I think that makes more sense. I was just using what was there before.

tuomas2:
I chose a bit different approach to do this, similar in Download documents activity, we index entities in FTS4 table. 1db398d0c6d19668fc8e0541ee794c9a2163e7ff

Issue #2684, opened 2023-07-21
Translator's context for 5.0


tuomas2:
Checklist (as suggested by @timbze ) from last release

- Czech @davidhrbac
- German
- Russian 
- Finnish @tuomas2 
- French
- Hungarian @krisek
- Portuguese
- Portuguese (Brazilian) @pepaulo 
- Afrikaans
- India (Hindi) @kesarsauce 
- South Korea 
- Lithuanian
- Spanish @timbze 
- Polish
- Arabic
- Italian @LAfricain
- Slovenian
- Chinese (CN&TW)
- Ukrainian
- Serbian (latin & rs)
- Kazakh
- Esperanto @verdulo 
- Norweigian
- Turkish @alerque 
- Estonian
- Azerbaijani 
- Dutch @marnix
- Slovak
- Uzbek
- Kyrgyz

[Excel version in Google Sheets](
https://docs.google.com/spreadsheets/d/1NHe7YtnRQ1lwEXh7Y3hf3V7CmsUirr6wIvsUX-sADeA/edit#gid=0)

Related: 
- #1028

tuomas2:
Sent announcement to translators via Transifex announcements

> New AndBible release 5.0 is planned within the coming few months, during the autumn most likely. There are some major new features in the new release:
> 
> - Cloud synchronize (via Google drive currently)
> - Discrete mode in main app and separate “calculator” app
> - Workspace colors
> - Chained links windows
> - Synchronize groups
> - Search / filter labels
> - Backup Database: choose what to restore
> - MyBible / MySword modules,  Installing, uninstalling, Backup & restore
> - Custom repositories
> - Deep links 
> 
> Some of these features require new strings to be translated. Amount of new strings is, however, not very big, like last time. 
> 
> Before starting translating, please see video intro to new features; I believe that it serves best as context for this release: https://www.youtube.com/watch?v=XdKmoe95KhI. See also translator's context ticket, in which you can also send questions and comments if they arise: https://github.com/AndBible/and-bible/issues/2684 
> 
> As always, translator's instructions can be read from our wiki page: https://github.com/AndBible/and-bible/wiki/Translating-User-Interface
> 
> As we are in Beta phase, we are mainly fixing bugs and smoothing the rough edges. Although there can be still minor changes to strings, it is good time to start updating translations.
> 
> Thank you for your continued support and co-operation for the AndBible Open Source project!
> 
> Best regards & God bless,
> Tuomas
> Lead developer

[...]

PR #2915, opened 2023-10-12
Complete all missing Turkish translations

alerque:
c.f. #2684

### DIFF
--- a/app/bibleview-js/src/lang/tr.yaml
+++ b/app/bibleview-js/src/lang/tr.yaml
@@ -1,7 +1,7 @@
 # Open in "My Notes" title. (just demonstrating developer note)
 bookmarks: Yer İmleri
 bookmarksAndNotes: Yer İmleri ve Notlarım
-openMyNotes: "Notlarım'da Aç"
+openMyNotes: Notlarım'da Aç
 openStudyPad: Çalışma Defterinde Aç (%s)
 noteText: Dipnotlar (%s)
 noteTextWithoutType: Dipnotlar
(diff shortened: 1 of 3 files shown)


alerque:
I'm headed out for something else and have no idea when I'll be back at it. Obviously I tackled the two smaller resource files before digging into the big one! Feel free to poke if I forget about this. I'm leaving it as draft because I do plan to work on the strings.xml file too, but also feel free to merge and feed this back into Transifex at any time.

One disadvantage of working this way is some of the meta data like character count limits is not enforced. I tried to look up that info on the ones I thought might have restrictions, but one thing we probably do want to do is round-trip these through Transifex and check the diff. If that cuts some of them off I can go back and fix those.

tuomas2:
There are not many strings that have enforcements in use; I think only some google play related strings have limits.

tuomas2:
I'll leave this PR open until you are finished or when we are releasing. Or if you want to test translations in a beta build, just ping me to merge it and release in next beta build.

alerque:
This should now bring Turkish up to 100% translated. The strings here are of various quality, some of them could use some copy editing and all of them could use some consistency checking. I noticed the existing strings had several major inconsistencies and did not try to resolve those when translating new strings.

I think I'd like to see this first pass merged earlier rather than later, then round-tripped into Transifex (where other reviewers are more likely to be able to chip in than they are here) and then back to the code so we can make sure the import/export doesn't surface any errors like too-long strings, mismatched placeholders, etc. I'll try to take another pass at editing everything again later, but seeing it in a beta might be useful too.

alerque:
I discovered the API and CLI tooling is something I can use too, so I've taken care of the round-tripping into Transifex, so the issues with that are already fixed in this branch and Transifex is up to date.

It would still be nice to merge this PR for the Git history even though they would come through later in a regular translations pull anyway.

tuomas2:
So you have enough permissions to upload you translations to transifex? 

I can merge and trigger a build for you so you can test it.

tuomas2:
Test build will appear here after CI is finishing: https://github.com/AndBible/and-bible/releases/tag/build-736-turkish-translation

alerque:
> So you have enough permissions to upload you translations to transifex?

Yes. I was a little bit surprised too, but I guess because I'm marked as the coordinator for the Turkish team I have special privs. Or maybe anybody on the team can use the API, I don't know. In any case I was pleasantly surprised by how smoothly it went, a few oddities notwithstanding.

tuomas2:
> > So you have enough permissions to upload you translations to transifex?
> 
> Yes. I was a little bit surprised too, but I guess because I'm marked as the coordinator for the Turkish team I have special privs. Or maybe anybody on the team can use the API, I don't know. In any case I was pleasantly surprised by how smoothly it went, a few oddities notwithstanding.

I believe it is because you are coordinator. Normal translators can (as far as I know) only translate untranslated strings (or those that they have translated themselves).

tuomas2:
So everything is pushed there so it's fine to overwrite these files in git next time when I pull?

tuomas2:
Big thanks BTW for updating turkish :-)

alerque:
> So everything is pushed there so it's fine to overwrite these files in git next time when I pull?

Yes, exactly, I've already round-tripped all three resources, the online copies are the freshest again so other contributors can jump in there, and you can feel free to pull & force pull from Transifex again at any time. I may do some more work directly on the Git side but I can make sure I freshen the resources from Transifex before I start next time so I don't clobber anyone's contributions.

Thanks for facilitating this.

PR #2917, opened 2023-10-13
Normalize references to Strong's Concordance

alerque:
James Strong was a person and the concordance title is a possessive based on his name.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -153,7 +153,7 @@
     <string name="prefs_navigate_to_verse_title">Navigate to verse</string>
     <string name="prefs_navigate_to_verse_summary">Choose verse (and chapter) when selecting a passage</string>
     <string name="prefs_open_links_in_special_window_title">Links window</string>
-    <string name="prefs_open_links_in_special_window_summary">Open links in special window, for quicker display of cross-references and Strongs</string>
+    <string name="prefs_open_links_in_special_window_summary">Open links in special window, for quicker display of cross-references and Strong\'s</string>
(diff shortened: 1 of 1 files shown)


alerque:
I noticed while trying to clean up the Turkish translation of these that the source strings were somewhat inconsistent.

tuomas2:
We are using british english accent in Transifex for minor corrections for english translation so that transifex does not invalidate all related translations. 

I will make the fix there (if it's not already fixed).

alerque:
Hmm, interesting arrangement. I can see not wanting no do too much invalidation, but if what I saw people doing in TR is any indication, the inconsistent source strings may be causing even worse translations. I don't know how widespread that will be in this case, but maybe I'll look through a few other languages and see if I can make sense of anything along those lines.

BTW, wouldn't the lead up to a major new release be the *least disruptive* time to fine tune source strings?

tuomas2:
> BTW, wouldn't the lead up to a major new release be the _least disruptive_ time to fine tune source strings?

In principle yes, but not at this stage (less than 3 weeks to release) :-)

alerque:
Okay, I wasn't sure what the timeline was. Personally my interest was more in having consistent sources to translate *from* than the English app experience, but of course I have different priorities here ;-) If you don't want to merge this I understand, but maybe I'll keep the branch around for other potential source text issues I run across and you or whoever can choose between cherry picking stuff to apply to sources, saving some bits to apply later at a more convenient time in the life cycle, discarding stuff you don't want at all, or using a temporary work around vis-à-vis the British shindig.

tuomas2:
Fixed in transifex

PR #2918, opened 2023-10-13
Drop incomplete translation warning from Turkish, resync

alerque:
I've been playing around with the app build you made a few hours ago and so far it doesn't seem to bad. I'll keep working on copy-editing as I get chance(s), but I think it's fair to drop the nag screen that comes up on updates now.

Also unrelated and I din't make the change, but I see in transifex that *cs* and *uk* translations are marked in Transifex as 'ready for use'. Is there a reason those 100% complete ones shouldn't be white listed yet? I can append that change to this PR too if desired.

### DIFF
--- a/app/src/main/java/net/bible/service/common/CommonUtils.kt
+++ b/app/src/main/java/net/bible/service/common/CommonUtils.kt
@@ -1368,8 +1368,8 @@ object CommonUtils : CommonUtilsBase() {
 
         Log.i(TAG, "Language tag $languageTag, code $languageCode")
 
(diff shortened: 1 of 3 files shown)


tuomas2:
Thanks. I have #2883 reminding me to update the good translations list. 
I'm having make tx-pull cmd which also compiles play descriptions, but I can run it later myself.

tuomas2:
Of course you can help me by checking new good translations and making pr :-) But new good translations hopefully appearing with good speed upcoming weeks (hopefully), so not checking that every day.

PR #2919, opened 2023-10-13
Use SVG logo in README for nicer GitHub dark theme views

alerque:
This is a minor detail, but it has bugged me for a while ever time I open the project on GitHub. GitHub has had an official dark mode for a while, and logos with white backgrounds don't play very nice any more.

This was probably not supported when the logo was first added, but in 2022 they unblocked SVGs as images in Markdown. The PNG could have been made transparent as well, but using the SVG resource already in the repository seems to make the most sense to me. The only caveat is if this README is ever rendered anywhere *off of GitHub* that does not support SVG images. If that's the case then we'd want to post a transparent PNG instead.

Here is a before/after:

20231013_17h01m50s_grim


tuomas2:
Thanks!

PR #2940, opened 2023-10-19
Translate a handful of TR strings that got copied from EN

alerque:
In spite of 100% coverage on Transifex I noticed a few English strings in my beta build. It turns out the mistake was probably mine, I copied some blocks around while making sure the TR was complete with everything in the EN, and a few strings did not get subsequently translated.

Also in poking around the app I was not comfortable with the 'purchase developer time' language. Honestly it strikes me a bit off in English too, I would suggest something like 'sponsor developer time' or even 'pay for', but 'purchase' is more related to a product. In Turkish the distinction is even more pronounced. I changed the jargon to 'by sponsoring development', 'become a sponsor', etc. as appropriate for each context.


tuomas2:
> Also in poking around the app I was not comfortable with the 'purchase developer time' language. Honestly it strikes me a bit off in English too, I would suggest something like 'sponsor developer time' or even 'pay for', but 'purchase' is more related to a product. 

Changing English to "Sponsor app development". Thank you for suggestion!

I try to avoid speaking of donation which is without compensation. I'm not receiving donations but selling development time. Sponsoring sounds to me to be the correct term.

PR #2967, opened 2023-10-28
Added line breaks and 'bullets' to help text

agrogers:


tuomas2:
Original english strings may not be changed this way (would invalidate translations). We could do this change in en-GB (via transifex), however.

PR #2978, opened 2023-10-30
Disable Inline References option when can't be shown

agrogers:
It was a little confusing that I could have Inline Reference enabled but they would not display. This change disables the Inline option if Cross References is not enabled.

image


tuomas2:
Looks good. Thanks!

Issue #2945, opened 2023-10-21
F-droid metadata needs changes for AndBible 5.0


tuomas2:
variant configuration and possibly node upgrade.

tuomas2:
Added tag v5.0.740-fdroid so we will get it try to build that release candidate version. Then we can do PR to f-droid metadata to fix things there.

linsui:
https://github.com/AndBible/and-bible/blob/5cb9cd9550e2386546677fb3a7219d9d0a781eef/app/build.gradle.kts#L356

These libs are non-free.

tuomas2:
> https://github.com/AndBible/and-bible/blob/5cb9cd9550e2386546677fb3a7219d9d0a781eef/app/build.gradle.kts#L356
> 
> These libs are non-free.

are you sure, did you check the license?

linsui:
Emmm, everything in GMS is non-free. Anyway, you can check it at https://maven.google.com/web/index.html#com.google.android.gms:play-services-auth:20.7.0.

tuomas2:
Do you think F-droid does not allow such? If it does not we have to disable sync feature from f-droid builds.

tuomas2:
License is this https://developer.android.com/studio/terms

linsui:
As a core contributor of F-Droid I'm sure GMS libs are non-free and not allowed by F-Droid. :) I found this issue because the build failed. https://gitlab.com/linsui/fdroiddata/-/jobs/5408600494#L604

tuomas2:
Ah, I see. Thanks for finding your way here :-) Then I will remove Google Drive support from F-droid builds. That's a pity though, do you know if there's any "allowed" way to use Google Drive API?

What also needs to be changed probably is to 
- upgrade node (not sure if it is needed but I'm using now myself node version 20) and
- change build to use standardFdroidRelease variant (instead of fdroidRelease if I remember correctly).

tuomas2:
I will do the necessary changes (related to google drive) next week if I have time and trigger new f-droid build then.

linsui:
> That's a pity though, do you know if there's any "allowed" way to use Google Drive API?

I thought you can sync the files to a local directory so that the user can use whatever they want to sync the data.

When the new version is ready just ping me here. :) I'll update the metadata.

tuomas2:
Disabled Google Drive device syncing feature and pushed new tag v5.0.741-fdroid @linsui

tuomas2:
[...]

Issue #2615, opened 2023-06-10
Implement NextCloud interface to Device Synchronization


tuomas2:
We need a self-hosted option to enable synchronization.

Essentially need to implement [CloudAdapter](https://github.com/AndBible/and-bible/blob/develop/app/src/main/java/net/bible/service/cloudsync/CloudAdapter.kt) for NextCloud.

This is an straight-forward task even to someone who does not know AndBible codebase.

tuomas2:
Another option is to implement WebDav. I'm just thinking that NextCloud API might be easier to work with.

[NextCloud Android client documentation](https://docs.nextcloud.com/server/latest/developer_manual/client_apis/android_library/index.html#android-nextcloud-client-development)

https://github.com/nextcloud/android-library

tuomas2:
Added some strings to codebase so that we can implement this post-release, if not earlier.

tuomas2:
Estimate: 1-5 days. Or if we are lucky maybe just several hours. Will make better estimate if someone is seriously interested.

tuomas2:
Work started

meichthys:
I think we can close this issue now that it's implemented! Thanks @tuomas2 !
One improvement that would be nice would be if it was possible to configure the path on Nextcloud where the synchronization files were saved. Currently the files are saved in the root of Nextcloud which can cause the Nextcloud root directory to get somewhat cluttered.

ildar:
Another feature: login through the NC account provider (app Nextcloud)
without manually creating a token and logging in. One working example is
https://f-droid.org/en/packages/it.niedermann.owncloud.notes

tuomas2:
You can create new tickets for the improvements, closing this.

PR #2981, opened 2023-11-01
F-droid fixes

tuomas2:
Related #2945

Remove non-free stuff (google drive related) from F-droid build

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -351,16 +351,17 @@ dependencies {
     implementation("net.objecthunter:exp4j:0.4.8")
     implementation("com.github.requery:sqlite-android:$sqliteAndroidVersion")
 
-
-    // Google Drive API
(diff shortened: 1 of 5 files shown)


tuomas2:
Still build fails for F-droid build because GoogleDriveCloudAdapter.kt is built (even though it is not used). Should somehow exclude that from being built.

ildar:
Thanks for fixing this! F-Droid becomes even more important these days.

tuomas2:
It's just pity that Google Drive support can't be enabled in F-droid build.

ildar:
On Thu, Nov 2, 2023 at 3:41 PM Tuomas Airaksinen wrote:

> It's just pity that Google Drive support can't be enabled in F-droid build.
>
It may be convenient but not essential after all.
thanks!

tuomas2:
#2615  will fix that

PR #3063, opened 2023-11-24
Original language icon improvements

agrogers:
Adds an icon for each display state and reduces the file size of the original icons.

image

image


tuomas2:
Please do any future smallish prs on top of current-stable so we can release them quicker.

Nothing yet in develop so this can be merged directly to current-stable, even though it is on top of current-stable.

agrogers:
OK

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

PR #3064, opened 2023-11-24
Add a Strongs "Hidden links" option

agrogers:
Something that bothers me is that when I show the Strongs links (without text) I get the dotted underline on almost every word which messes with the bookmarked verse underline and also makes all the text extra busy and not so nice to read.

This PR adds a fourth option to the Strongs Links.
image

At present it  functions exactly the same as the "Links" options but without showing the dotted underline.

However I want the single tap option to operate a little differently. It is fine for verses with bookmarks... the verse action dialog is displayed allowing me to add a new bookmark or view the Strongs link. But I want the dialog to **always display** when the links are hidden, even if the verse is not bookmarked. It would essentially work as though their are no links and so show the dialog but **if there is a linked word** it will display 'Definition & Morphology' button.

image

I can't find the code that does the checking for an existing bookmark. before opening (or not opening) the ambiguous dialog. I think all I need to do is check if we are using Hidden Links and if we are always show the dialog rather than open the Strongs link in the links window.

Could you point me to where this check takes place?
Thanks.

### DIFF
--- a/app/bibleview-js/src/components/OSIS/W.vue
+++ b/app/bibleview-js/src/components/OSIS/W.vue
@@ -101,7 +101,7 @@ function goToLink(event: MouseEvent, url: string) {
         resetHighlights();
         isHighlighted.value = true;
         addCustom(() => isHighlighted.value = false);
-    }, {priority, icon: "custom-morph", title: strings.strongsAndMorph, dottedStrongs: !showStrongsSeparately.value});
+    }, {priority, icon: "custom-morph", title: strings.strongsAndMorph, dottedStrongs: !showStrongsSeparately.value, hiddenStrongs:showStrongsHidden.value});
 }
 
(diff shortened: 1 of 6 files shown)


tuomas2:
AmbiguousSelection.vue:handle and W.vue: goToLink. Options in addEventFunction are readable in handle function.

agrogers:
Thanks. I will check it out.

___________________________
***@***.***
(+95) 09 250018669 (Myanmar)

agrogers:
It now works how I wanted it... no dotted underline (hidden strongs mode) always shows the dialog as it would normally when there is no dotted underline and shows the morphology button in the action dialog.

tuomas2 (approved):
Looks good, merging.

tuomas2:
After small testing works very well. Thanks!

PR #3080, opened 2023-11-30
Search epubs

tuomas2:
Implement #3051 

Still todo:

- [x] indexing status display
- [x] check that indexing works OK
- [ ] all around testing
- [ ] add context to search items? (surrounding bvas)
- [ ] ~open results in a window~
- [x] db in another file, saved with mod structure
- [ ] going back from search results, should not reset the form
- [x] should probably save the setting


tuomas2:
I was impressed that FTS5 gives highlighting for free.

PR #3279, opened 2024-06-23
"Find" on selection text: still submit selections greater than 15 chars

wesinator:
This PR makes it so that the reader can still select text greater than 15 characters in length and send it to the search.
Currently, selections greater than 15 don't generate a "Find" menu option because it always includes the search term in the menu title, and excludes a selection longer than 15.
This is fixed by simply using the shorthand "Find" from the "search" string resource if the selected text is longer than 15 chars.

[NO_TRAIN]::


tuomas2:
Also looks good to me. Let's try...

tuomas2:
There were some compile and other issues, which are fixed in 190d08f4c77dc23c50d6918b5b006cbb15ef3097. I am wondering did you manage to test this at all if it didn't even compile.

tuomas2:
But functionalities are looking useful and good. Thank you!

PR #3354, opened 2024-12-31
Better e-ink support

### LINKED ISSUE #282: Page flip option for E-Ink devices
From user via email:

> Can you add a page flip option for your Bible app to be more accommodating for those wanting to use this app from an Android E-Ink device? 
> 
> In order to be (properly) usable in an E-Ink device, page flip would have to be added as a selectable option to use instead of continuous scrolling. The page flip would need to have either no animations, or a "fade out/fade in" type of animation. Any sliding/moving animations (including the present continuous scrolling) create "ghosting" with E-Ink screens, and do not display well. Compatibility with any types of animations is very limited on E-Ink devices, and can make the user experience impractical.
> 
> If you could add an option for page flip in the future, this would allow your app to be usable on E-Ink devices. 
> So far, the only apps that are usable for Bible reading are Android Kindle and Kobo apps, which have an option to disable animations on page flips. Unfortunately, this means that I cannot effectively do Bible study on my e-reader.
> 
> If you have any questions about my suggestion, feel free to let me know.
>

### LINKED ISSUE #842: Make option to disable smooth scrolling / animations
Via support email:
> Hello team,
> Thank you for your time and many efforts this is a very useful & nice app!
> 
> As I intentinally turned off any menu animation and any smooth scrolling (@developer's options), pls. render it possible to deactivate this as option.
> 
> Especially when using windows and fast scrolling the depending windows follow very slowly coasting till stop, what delays significantly - it was earlier not like this!
> 
> I would appreciate an update in this regard. Thx.
> 
>

### LINKED ISSUE #1126: Make And Bible work better with e-ink devices
> Yes, there are E-Ink devices based on Android. I personally use the ONYX 
BOOX Note Pro with Android 9 installed. As I have the AndBible app 
installed on my mobile phone, I am basically already used to the design and 
would love to be able to use it on the E-Ink tablet as well (as this is the 
primary device for my Bible studies). For now, the contrast of the top 
navigation bar is so low, that the icons can't be seen at all. That's by 
now means a complaint, just an idea on how to improve the app or make it 
more "versatile". I really like the app, and I am very thankful for it - it 
is truly a blessing!

https://support.andbible.org/scp/tickets.php?id=561

### LINKED ISSUE #3351: e-ink should not assume monochrome
I've been seeing some tagged test releases with setting adjustments targeting e-ink displays. This is exciting and something I've looked forward to for a while. I haven't had a chance to update my tablet(s) and give it a test drive yet, but I plan to.

In the mean time just looking at the code branch I thought I'd point out the setting seems to be conflating two things. Not all e-ink displays are monochrome. In fact there are quite a few color options out there. I don't happen to have one yet, but there are more and more out there. I think it's a bit short sighted to assume monochrome == eink. Most of the adjustments probably apply to both, but

* the *biggest* common issue is probably refresh rate and the clumsiness of scrolling (c.f. #282).
* secondarily I think almost all eink displays to date have issues with contrast. Even the color ones are not vibrant and while they have great contrast for reading they don't tend to do mid-tones well. Obviously the monochrome design language, ditching shadows and transition effects, etc. will benefit even color displays because of the contrast and refresh rate
* finally of course many of them *are* only grey scale.

In naming the setting and grouping UI changes into logical groups I would probably target the e-ink specific issues with more than one setting so that the folks with color screens can actually use them while benefiting from the other UI changes.

tuomas2:
Addresses various issues (see issues tagged with [E-ink tag](https://github.com/AndBible/and-bible/labels/E-ink))

Closes #1126
Closes #3351 
Closes #842
Closes #282

### DIFF
--- a/app/bibleview-js/src/components/BibleView.vue
+++ b/app/bibleview-js/src/components/BibleView.vue
@@ -16,7 +16,7 @@
   -->
 
 <template>
-  <div @click="ambiguousSelection?.handle" :class="{night: appSettings.nightMode}" :style="topStyle" :dir="direction">
+  <div @click="ambiguousSelection?.handle" :class="{night: appSettings.nightMode, noAnimation}" :style="topStyle" :dir="direction">
     <div class="background" :style="backgroundStyle"/>
     <div :style="`height:${calculatedConfig.topOffset}px`"/>
     <div :style="modalStyle" id="modals"/>
@@ -321,6 +321,9 @@ setupEventBusListener("scroll_up", () => scrollUpDown(true));
 
 useSharing({topElement, android});
 const direction = computed(() => appSettings.rightToLeft ? "rtl" : "ltr");
+
+const noAnimation = computed(() => appSettings.monochromeMode);
+
 </script>
 <style lang="scss">
 @import "~@/common.scss";
--- a/app/bibleview-js/src/components/BibleView.vue
+++ b/app/bibleview-js/src/components/BibleView.vue
@@ -28,6 +28,7 @@
     <div class="window-id" v-if="appSettings.errorBox">{{appSettings.windowId}}</div>
     <DevelopmentMode :current-verse="currentVerse" v-if="config.developmentMode"/>
     <div v-if="calculatedConfig.topMargin > 0" class="top-margin" :style="`height: ${calculatedConfig.topOffset}px;`"/>
+    <div v-if="noAnimation && calculatedConfig.topMargin > 0" class="bottom-margin" :style="`height: ${calculatedConfig.topOffset}px;`"/>
     <div v-if="appSettings.hasActiveIndicator">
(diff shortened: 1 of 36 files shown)


app/bibleview-js/src/components/BibleView.vue
```diff
@@ -342,6 +365,9 @@ $ring-color: $button-grey;
     border: $ring-thickness solid $ring-color;
     border-radius: 50%;
     animation: lds-ring 1.2s cubic-bezier(0.5, 0, 0.5, 1) infinite;
+    .noAnimation & {
+      animation: none; // TODO: better loading indicator for noAnimation
```
tuomas2: - [x] TODO

tuomas2:
Note: this work was sponsored by me.

PR #3370, opened 2025-01-16
Nextcloud implementation for Device Sync

### LINKED ISSUE #2615: Implement NextCloud interface to Device Synchronization

tuomas2:
Closes  #2615


app/src/main/java/net/bible/service/cloudsync/nextcloud/NextCloudAdapter.kt
```diff
@@ -0,0 +1,230 @@
...
+            return@withContext true
+        } catch (e: Exception) {
+            Log.e(TAG, "Login to NextCloud failed", e)
+            _client = null
+            return@withContext false
+        }
+    }
+
+    override suspend fun signOut() {
+        _client = null
```
tuomas2, line 81: sign out should also remove configurations (getConfigs()). After signing out it must be again fresh start.

tuomas2: actually that is taken care of in CloudSync.signOut()

PR #3411, opened 2025-04-30
Feature/memorize

### LINKED ISSUE #2547: Bible Memory Feature
Hello!

Thank you for a wonderful tool!. I love the app.. recently I was wanting to try and memorize scripture more and while I found several tools that did a good job they weren't really open source or Foss which I have come to appreciate and look for in an app. Its also yet another app to install. I thought it might make sense to be incorporated into the app I already use for studying Gods word and thus this feature request..

I have been using the play store Verse Locker app by scripturememory.com.. I like the options it has for memorizing.. like fill in the blank, listen etc.. I feel like the cloud login and network component would be unnecessary if it were built into andbible so maybe there isnt to much extra stuff to add since the scripture is already present . I realize this might not be in scope but I figured I would put it out there. 

Thank you again!

tuomas2:
Closes #2547


app/src/main/java/net/bible/android/view/activity/settings/SettingsActivity.kt
```diff
@@ -175,13 +175,6 @@ class SettingsActivity: ActivityBase() {
 }
 
 class SettingsFragment : PreferenceFragmentCompat() {
-	override fun onDisplayPreferenceDialog(preference: Preference) {
-        if(parentFragmentManager.findFragmentByTag("customTag") != null)
-            return
-
-        super.onDisplayPreferenceDialog(preference)
-	}
```
tuomas2, line 183: - [x] check this change, probably not intentional

app/bibleview-js/src/lang/default.yaml
```diff
@@ -59,10 +59,18 @@ openDownloads: Go to Add-on downloads
 verseShare: Share
 verseSpeak: Speak
 verseCompare: Compare
+verseMemorize: Memorize
 verseNote: Note
 verseMyNotes: My Notes
 verseShareLong: Share this verse
 verseCompareLong: Compare verse
+verseMemorizeLong: Memorize verse
```
tuomas2:
```suggestion
verseMemorizeLong: Memorize verses
```

app/src/main/java/net/bible/android/view/activity/page/BibleView.kt
```diff
@@ -445,6 +450,12 @@ class BibleView(val mainBibleActivity: MainBibleActivity,
         linkControl.openCompare(verseRange)
     }
 
+    internal fun memorizeSelection(selection: Selection? = currentSelection) {
+        Log.i(TAG, "compareSelection")
```
tuomas2:
```suggestion
        Log.i(TAG, "memorizeSelection")
```

app/src/main/java/net/bible/android/view/activity/settings/SettingsActivity.kt
tuomas2: I think these lines have no use anymore, so let's keep it like this.

Issue #3420, opened 2025-05-17
Show andBible Version in App.


meichthys:
**Is your feature request related to a problem? 
When trying to file a bug report, it would be nice if it was a way to see the current app version inside the app.

**Describe the solution you'd like**
I'm thinking a small font at the very bottom of the main menu would be sufficient. Tapping on that version number could also copy all of the app debug details that are normally included in the debug email. ( Like the information I've copied below.)

**Additional context**
Device info:
App id: net.bible.android.activity
Version: 5.0.872#627f476 github standard release (built 08/05/25 07:06:51)
Android version: 15
Android SDK version: 35
Manufacturer: Google
Hardware: panther
Product: panther
Device: panther
Brand: google
Model: Pixel 7
Storage Mb free: 9463
WebView version: 136.0.7103.125.0
SQLITE version: 3.45.0
Used heap memory in Mb: 48
Max heap memory in Mb: 256

meichthys:
![Screenshot_20250517-083128.png](https://github.com/user-attachments/assets/048f2407-e734-4dcb-a42d-cca5fc26af78)

tuomas2:
Good idea. At the moment it can be found at the bottom of help screen.

meichthys:
Ah, okay, I hadn't seen it there before.

timbze:
One idea... Though it's a bit more than this issue is suggesting. Not sure if we want to add anything at the top, or just leave it with adding version at bottom.

<img width="486" height="471" alt="Image" src="https://github.com/user-attachments/assets/23f12a4f-9dee-468b-b86c-5b7d9c2c38cf" />

<img width="486" height="471" alt="Image" src="https://github.com/user-attachments/assets/56b5b71c-9892-47fc-b218-244461420a4f" />

timbze:
On tablet the margins are off:

<img width="519" height="355" alt="Image" src="https://github.com/user-attachments/assets/3ce5933f-8487-4683-b8c5-98075cab8e44" />

timbze:
> On tablet the margins are off:

On my phone too. I guess changing the layout_height to wrap may have caused this.

timbze:
Fixed by #3453 #3456

PR #3456, opened 2025-08-04
Margins #3453 #3420

timbze:
Fixes for nav drawer logo margins.
#3453 
#3420

Also removed debug/values-en/strings.xml which is not needed. There is the debug/values/strings.xml which has correct values.


app/src/debug/res/values-en/strings.xml
```diff
@@ -1,22 +0,0 @@
...
-  ~ See the GNU General Public License for more details.
-  ~
-  ~ You should have received a copy of the GNU General Public License along with AndBible.
-  ~ If not, see http://www.gnu.org/licenses/.
-  -->
-
-<resources>
-    <string name="app_name_medium">Bible Study App (And Bible) *Debug*</string>
-    <string name="app_name_long">Bible Study App, by And Bible Open Source Project *Debug*</string>
-</resources>
```
tuomas2, line 22: why this?

timbze: It's out of date, medium name is now `Bible Study (AndBible) *Debug*`. This file already serves the purpose and when I remove the other file this is now used: https://github.com/AndBible/and-bible/blob/develop/app/src/debug/res/values/strings.xml

timbze: I realized the issue because of the space in "And Bible" in my dev build

PR #3555, opened 2025-11-06
Bring jsword as a submodule back (?)

tuomas2:
I'm not sure if we should use jsword as a submodule still. Let's undo it for now. But leave PR to easily revert again...


tuomas2:
Re-implemented again

PR #3564, opened 2025-11-19
Add files via upload

jadsongmatos:


tuomas2:
Did you create this yourself?

jadsongmatos:
> Did you create this yourself?

yes

timbze:
This in itself does not add the file to AB. There's a map object someplace that needs to be updated. At least that.

However, I would not add it to AB because it includes apocryphal books. Everyone can use it for themselves if they have the file.

tuomas2:
@timbze and isn't it our general policy not to add more reading plans but rather distribute reading plan [add-on modules](https://github.com/AndBible/and-bible/wiki/Reading-Plans) via [modules repository](https://github.com/AndBible/data)?

timbze:
> @timbze and isn't it our general policy not to add more reading plans but rather distribute reading plan [add-on modules](https://github.com/AndBible/and-bible/wiki/Reading-Plans) via [modules repository](https://github.com/AndBible/data)?

We have said that before yes, but in my view it had more to do with lifting up other ideologies which we didn't want to do.

PR #3565, opened 2025-11-22
Bugfix/#1848 tts adds history item

### LINKED ISSUE #1848: Speak makes history item for each verse
**Describe the bug**
Only happens when "Text follows speech" is enabled.

Start speak, and then look at history after some verses are read. Each verse gets added to history.

https://support.andbible.org/scp/tickets.php?id=923

Build 612 (still exists in 834)

Tritonfc:
Fixes #1848 which adds a history item for every scrolled page during TTS


Tritonfc:
Created this branch from my other PR so it has the commits from those too. A small mistake. Will always create future branches exclusively from `current_stable` moving forward

tuomas2 (approved):
Looks good. Let's give it a try.

Issue #3266, opened 2024-06-10
Eq issue


Noitarud:
Maybe this was deliberate, or not, but if I enter 1+1 and repeatedly press = it will eventually enter. Maybe:
- disable entry if having to do a sum
- enter based on whole keyed string, even if more complex like "6/7.0="

Ver 5 0 735

tuomas2:
also could make calculations deliberately a bit slower, like 0.5 sec per calc.

Noitarud:
Or maybe 101 limit also, you generally get distracted by then. Maybe check the second last key and if that was not number then fail before checking if correct.

Tritonfc:
Was this fixed? Doing a sum and repeatedly pressing enter just keeps adding the numbers but doesnt enter the app

tuomas2:
No there is no fix. I would assume it will enter the app after number has incremented to the actual pin number? If you set up pin number 3 for example, you ned to press = key 2 times.

PR #3579, opened 2025-12-20
Bugfix/#3266 calculator equals issue

Tritonfc:
This fixes issue #3266 where pressing equals on the calculator after previous operation access the app. The fix basically disables validating the pin when doing subsequent operations and only validates on direct pin entry.


tuomas2 (approved):
Looks good to me

PR #3593, opened 2026-01-09
fix(bibleview-js): kjv idiosyncratic Milestone type x-p

loosestrife:
the default kjv downloaded didnt have paragraphs because the Milestone type was "x-p" instead of "line"

**Screenshots**
the pilcrow is messed up and the verse number is in the wrong place though, plz advz
image


tuomas2 (commented):
# Claude's Code Review: fix(bibleview-js): kjv idiosyncratic Milestone type x-p

## Overview

This PR aims to fix paragraph rendering for the KJV Bible module by handling the `x-p` Milestone type. However, it also includes several unrelated changes that significantly expand the scope.

## Summary of Changes

1. **Core Fix**: Handle `x-p` Milestone type for KJV paragraphs
2. **Logger System**: New logging utility class
3. **ProcessTextActivity**: New Android activity for "Cite to Quote" feature
4. **Build Config Changes**: Disabled minification, added new dependency
5. **Various Bug Fixes**: Null safety improvements in StudyPadDocument
6. **Test Data Changes**: Modified development test data

---

## Critical Issues ⚠️

### 1. PR Scope Too Broad
This PR mixes multiple unrelated features:
- Milestone `x-p` fix
- New Logger system
- ProcessTextActivity (entirely new feature)
- Build configuration changes

**Recommendation**: Split into separate PRs for easier review and safer merging.

### 2. New Dependency Added Without Justification
```json
"browser-util-inspect": "~0.2.0",
```
- This package adds object inspection in the browser
- The package has only ~200 weekly downloads, raising maintenance concerns
- Consider if `JSON.stringify()` with a replacer would suffice

### 3. Minification Disabled in Production Build
```typescript
// vite.config.mts
minify: false, // no need to minify theres no network download and this is open source
```
**Issues**:
- Increases bundle size unnecessarily
- Open source doesn't mean users shouldn't benefit from smaller bundles
- WebView still loads these files, affecting memory and parse time
- **Recommendation**: Revert this change

### 4. Incomplete ProcessTextActivity Implementation
```kotlin
val theRawQuote = theBible.getRawText(theKey);
// val theQuote = preprocess(theRawQuote, html=false, linebreaks=true, pilcrows=true, verseNumbers="\${i}");
val theQuote = theRawQuote;
val out = theQuote + theCite;
```
- Commented-out preprocessing code
- Raw OSIS/XML likely returned instead of plain text
- Missing string resource for "No Bible selected" and other user-facing messages
- Feature seems incomplete

---

## Code Quality Issues

### 1. Logger Implementation Has Issues
```typescript
// logger.ts
_log(method: "warn"|"log"|"debug", ...data: any[]){
    // ...
    if(runningInAndroid){
        data = data.map(x => typeof(x) == "object" ? inspect(x) : x);
        const appSettings = inject(appSettingsKey);  // ❌ inject() called outside setup()
        if (this.documentWarningActive && !appSettings?.errorBox) {
            //emitThis = false;
        }
    }
}
```
- `inject()` is called inside a method, but Vue's `inject()` can only be used in setup context
- Commented-out logic suggests incomplete implementation

### 2. Missing Newlines at End of Files
Multiple files now missing trailing newlines:
- `Milestone.vue`
- `Verse.vue`
- `testdata.js`
- `logger.ts`

### 3. Kotlin Style Inconsistencies
```kotlin
// ProcessTextActivity.kt
import android.widget.Toast;  // Semicolon (Java style, not Kotlin)
val theCite = selectedText.toString();  // Semicolons throughout
```
**Recommendation**: Remove unnecessary semicolons to follow Kotlin conventions.

---

## Good Changes ✅

### 1. Core Fix is Correct
```typescript
// Milestone.vue
checkUnsupportedProps(props, "type", ["x-strongsMarkup", "x-PN", "line", "x-p"]);
const paragraphBreak = computed(() => props.type === "line" || props.type === 'x-p');
```
The actual fix for KJV paragraphs is simple and correct.

### 2. Null Safety Improvements
```typescript
// StudyPadDocument.vue
updateBookmarkToLabels(...(genericBookmarkToLabels_ ?? []))
globalBookmarks.updateBookmarks(genericBookmarks ?? []);
```
Good defensive coding.

### 3. Guard for Non-Android Environment
```typescript
// android.ts
const runningInAndroid = !!window.android;
function setClientReady() {
    if(runningInAndroid) {
        window.android.setClientReady();
    }
}
```
Good for development mode in browser.

---

## Test Coverage

- No new tests added for the Milestone `x-p` type change
- No tests for the new Logger class
- No tests for ProcessTextActivity
- **Recommendation**: Add at least a unit test verifying `x-p` triggers paragraph break

---

## Recommendations

1. **Split this PR** into at least 3 separate PRs:
   - Milestone `x-p` fix (the actual bug fix)
   - Logger system (if needed)
   - ProcessTextActivity (new feature)

2. **Revert `minify: false`** - this provides no benefit and increases bundle size

3. **Reconsider the `browser-util-inspect` dependency** - is it necessary?

4. **Complete or remove ProcessTextActivity** - the feature appears incomplete

5. **Add tests** for the Milestone type change

6. **Fix the Logger's `inject()` usage** - it won't work outside Vue setup context

7. **Add missing trailing newlines** to modified files

8. **Clean up Kotlin semicolons** in ProcessTextActivity

---

## Verdict

The core fix (handling `x-p` Milestone type) is correct and simple. However, this PR introduces too many unrelated changes, some of which are incomplete or problematic. I recommend:

- Extract the Milestone fix into its own small PR for immediate merge
- Address the other changes separately after further discussion and completion

app/bibleview-js/src/components/BibleView.vue
```diff
@@ -209,22 +212,24 @@ const loadingCount = ref(0);
 
 function addDocuments(...docs: AnyDocument[]) {
     async function doAddDocuments() {
-        console.log("doAddDocuments, start")
+        logger.info("doAddDocuments, start")
```
tuomas2, line 215: why not console.log? see patchAndroidConsole (which makes sure js side logging is going to android side)

tuomas2:
## Additional issue: Pilcrow overlapping text

Looking at the screenshot in the PR, there's a visual bug where the pilcrow (¶) character overlaps with the text "There" below it:

The cause is in `common.scss`:
```css
.paragraphBreak {
  display: block;
  height: 0.5em;  /* constrains height but text overflows */
}
```

The marker text overflows and renders on top of the following content.

**Quick fixes:**
- Hide marker when paragraphBreak: `{{ paragraphBreak ? '' : marker }}`
- Or CSS: `overflow: hidden;` or `font-size: 0;`
- Or use `min-height` instead of `height`

---

## Broader concern about complexity

The fundamental issue is that OSIS XML has the `<milestone>` *inside* the verse, but visually the paragraph break should appear *before* the verse number. This is a structural mismatch between data format and desired rendering.

While the core fix (recognizing `type="x-p"`) is simple and correct, any solution that tries to reorder the paragraph break to appear before the verse number would require complex workarounds (CSS flexbox reordering, provide/inject patterns, or OSIS preprocessing).

**My preference:** I'd rather keep things simple. If the straightforward CSS fix (just showing a block-level break where the milestone appears) doesn't produce ideal results, I'm not sure it's worth adding complex "ad hoc" solutions for this edge case. 

What's your preference - keep it simple even if the visual result isn't perfect, or is there a clean solution I'm missing?

app/bibleview-js/src/components/BibleView.vue
loosestrife: i noticed that some document warnings were only emitted if there was an app setting for a debug window, so i wanted to have a mechanism to impose that, separate out document warnings, and get the filename added to the log message

app/src/main/java/net/bible/android/activity/ProcessTextActivity.kt
```diff
@@ -0,0 +1,83 @@
...
+                    Log.w(TAG, theMsg);
+                    Toast.makeText(this, theMsg, Toast.LENGTH_SHORT).show();
+                    return;
+                }
+                val theRawQuote = theBible.getRawText(theKey);
+                // val theQuote = preprocess(theRawQuote, html=false, linebreaks=true, pilcrows=true, verseNumbers="\${i}");
+                val theQuote = theRawQuote;
+                val out = theQuote + theCite;
+                // if not readonly, replace cite with quote, otherwise put quote in clipboard
+                if(!isReadOnly){
```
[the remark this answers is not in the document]
loosestrife: what i would want to do is put in "[Jn 3:16]" with brackets, so i would prefer only a space instead of a newline

loosestrife:
is there a way to get the text out of bibleview-js for the cite-to-quote feature or do i need to do some transformations in the kotlin?  if i need to do that processing in kotlin then that same code can write out good xml that the vue part can easily handle, and the document warnings can be moved to that code, though, vue 3 appears to have introduced new document warnings that cant be suppressed

tuomas2:
I did some investigation on the `x-p` milestone type:

**OSIS schema check:** According to the official OSIS 2.1.1 schema, `x-p` is NOT a valid milestone type. The `milestonePt` enumeration only allows: `column`, `cQuote`, `footer`, `halfLine`, `header`, `line`, `pb`, `screen`.

**JSword's handling:** However, I found that JSword's XSL files (simple.xsl, html5.xsl) already handle `milestone[@type = 'x-p']` with a comment explicitly calling it "the KJV paragraph marker". The XSL also notes that `osis2mod` (CrossWire's module creation tool) transforms `<p>` elements into `<div type="x-p">` during module creation.

So `x-p` is a **CrossWire/SWORD convention** rather than official OSIS - and JSword already supports it.

**My concern:** Since JSword's comments specifically call this "the KJV paragraph marker", it seems potentially KJV-specific. Before merging, it would be good to know:
- Is `x-p` used by other modules, or just KJV?
- If it's KJV-only, should we make the handling module-specific, or request CrossWire to update KJV to use standard `type="line"`?

I'm hesitant to add generic handling for what might be a single module's quirk, especially without tests verifying other modules aren't affected.

loosestrife:
...so i need to use xsl for the transformation to plain text for the cite to passage feature?  also how would a feature to get a selected cite in another app to open the Bible to that verse work?

loosestrife:
yeah handling this single modules quirk means adding preprocessing to hoist a line break out of the verse while maintaining a pilcrow inside the verse, either with an xsl or a kotlin program, then that same system can be used to handle any other quirks found, but stuff other than the preferred kjv arent important enough to have quirks

loosestrife:
image
heres what i have so far.  i want the verse number to jump the pilcrow but i dont know how to do that, but this is probably good enough.  its a css trick

loosestrife:
...but it doesnt solve the problem of stripping the html out of the raw text from the jsword getRawText in the cite2verse feature which means reading osis markup with kotlin or xsl, the question is what the andbible devs want in andbible to maintain long term

loosestrife:
superseded

PR #3615, opened 2026-01-22
Fix: TTS auto-scroll loses sync with audio when font size is increased

### LINKED ISSUE #3611: TTS auto-scroll loses sync with audio when font size is increased
## Description

When using TTS (text-to-speech) playback with automatic scrolling enabled, the on-screen text and highlighting become out of sync with the spoken audio if the font size is increased. The audio progresses correctly, but the visible text lags behind.

## Steps to reproduce

1. Open any Bible text in AndBible
2. Enable TTS playback with automatic scrolling
3. Increase the font size in settings
4. Start TTS playback
5. Observe the mismatch between spoken text and on-screen position

## Expected behavior

The highlighted text and on-screen position should stay synchronized with the spoken TTS output, regardless of font size.

## Actual behavior

- With default font size, synchronization works well
- After increasing font size, the on-screen text scrolls too slowly
- The display may still show the beginning of a verse while TTS is already reading the end
- Manual scrolling is required continuously to keep up

## Technical analysis

The issue appears to be in `app/bibleview-js/src/composables/verse-notifier.ts`. The viewport detection algorithm uses fixed multipliers to find the currently visible verse:

```typescript
let y = calculatedConfig.value.topOffset + lineHeight.value * 0.8;
// Scans downward in increments of: lineHeight * 0.5
```

When font size increases, verses occupy more vertical space, but the scanning increments don't scale proportionally to actual verse element heights. This causes the algorithm to "lag behind" - it detects the previous verse as still visible when TTS has already moved to the next one.

### Related files
- `app/bibleview-js/src/composables/verse-notifier.ts` - verse visibility detection (likely root cause)
- `app/bibleview-js/src/composables/scroll.ts` - scroll position calculation
- `app/bibleview-js/src/components/BibleView.vue` - lineHeight computation

## References

- Support ticket: https://support.andbible.org/scp/tickets.php?id=3130

Tritonfc:
Implements and fixes #3611 by increasing line height span and starting closer to the offset

### DIFF
--- a/app/bibleview-js/src/composables/verse-notifier.ts
+++ b/app/bibleview-js/src/composables/verse-notifier.ts
@@ -54,7 +54,9 @@ export function useVerseNotifier(
     // totally frozen during scrolling
     const onScroll = throttle(() => {
         if (isScrolling.value) return;
-        let y = calculatedConfig.value.topOffset + lineHeight.value * 0.8;
+        // Start closer to the top offset to accurately detect the current verse,
+        // especially with larger font sizes where verses span multiple lines.
+        let y = calculatedConfig.value.topOffset + lineHeight.value * 0.3;
 
         // Find element, starting from right
         let element: Nullable<HTMLElement>;
@@ -80,7 +82,8 @@ export function useVerseNotifier(
                         }
                     }
                 }
-                y += lineHeight.value * 0.5;
+                // Use larger increment to better handle multi-line verses with large fonts
+                y += lineHeight.value * 1.0;
             }
         }
     }, 50);


app/bibleview-js/src/composables/verse-notifier.ts
```diff
@@ -54,7 +54,9 @@ export function useVerseNotifier(
     // totally frozen during scrolling
     const onScroll = throttle(() => {
         if (isScrolling.value) return;
-        let y = calculatedConfig.value.topOffset + lineHeight.value * 0.8;
+        // Start closer to the top offset to accurately detect the current verse,
```
tuomas2, line 57: Comments are fine in general, but do not comment the change that is being made in pull request.  Instead give explanation to the number if necessary. I would leave explanation just to commit message. 

If you want to explain this change to the reviewer, you can add such comments about change being made for reviewer in pull request as review comments.

tuomas2 (approved):
I will merge this but remove the comments

PR #3621, opened 2026-01-27
Fixed Ordering issue with search results

### LINKED ISSUE #3602: Find results are not in Bible order
After the latest update that added multiple translation search, I've noticed that the search results are not in Bible book order.

Tritonfc:
Fixes #3602


tuomas2:
## Code Review: Fixed Ordering issue with search results

### Overview
This PR fixes issue #3602 where search results were not displaying in proper Bible book order after the multi-translation search feature was added. The fix changes the sorting logic from using a single `Verse.ordinal` property to a composite sort using book ordinal, chapter, and verse number separately.

### Changes Analysis

#### 1. Search Result Sorting Fix (`SearchControl.kt`)

**The Problem:**
The original code used `it.normalizedVerse.ordinal` for sorting, but `Verse.ordinal` in JSword represents a flat index across the entire versification which can differ between translations. When grouping results from multiple translations with different versifications, this causes incorrect ordering.

**The Solution:**
```kotlin
searchResults.mainSearchResults.sortWith(compareBy(
    { it.normalizedVerse.book.ordinal },
    { it.normalizedVerse.chapter },
    { it.normalizedVerse.verse }
))
```

This correctly sorts by:
1. Book ordinal (Genesis=0, Exodus=1, etc.)
2. Chapter number
3. Verse number

**Assessment:** ✅ Correct fix. The composite sort ensures consistent Bible ordering regardless of versification differences between translations.

#### 2. CI Workflow Change (`.github/workflows/run-tests.yml`)

Changed from:
```yaml
run: ./gradlew --console plain testStandardGoogleplayRelease
```
To:
```yaml
run: ./gradlew --console plain testStandardGoogleplayDebugUnitTest
```

**Assessment:** ⚠️ **Unclear if intentional.** This changes:
- From release variant to debug variant
- From broader test task to specifically unit tests only

This change should either:
- Be in a separate PR if intentional, or
- Be explained in the PR description if related to the fix

### Suggestions

1. **Code Duplication:** The sorting logic is repeated for both `mainSearchResults` and `otherSearchResults`. Consider extracting a common comparator:
   ```kotlin
   val verseComparator = compareBy<GroupedSearchResult>(
       { it.normalizedVerse.book.ordinal },
       { it.normalizedVerse.chapter },
       { it.normalizedVerse.verse }
   )
   searchResults.mainSearchResults.sortWith(verseComparator)
   searchResults.otherSearchResults.sortWith(verseComparator)
   ```

2. **Update the comment:** The comment on line 189 still says "Sort by verse ordinal" but the sorting logic has changed. It should be updated to reflect the new multi-key sort.

3. **CI Change Clarification:** Please clarify whether the workflow change is intentional and why. If it's unrelated to the search fix, it should be a separate PR.

### Potential Issues

- **None identified** for the core fix. The sorting approach is sound and follows standard Bible book ordering.

### Test Coverage

- No new tests added. Consider adding a unit test that verifies search results are returned in correct Bible order when searching across translations with different versifications.

### Summary

| Aspect | Status |
|--------|--------|
| Code correctness | ✅ Good |
| Project conventions | ✅ Follows Kotlin style |
| Performance | ✅ No impact (same O(n log n) sort) |
| Test coverage | ⚠️ Could use a test |
| Security | ✅ No concerns |

---

**From @tuomas2:** Fix looks good to me. CI workflow change requires explanation though and should be reverted if no specific reason for it.

tuomas2:
**Regarding the CI workflow change:**

I investigated this and the change is actually **required** after the AGP 9.0 migration. In AGP 9.0, unit test tasks only exist for Debug variants:

```
app:testStandardGoogleplayDebugUnitTest
app:testStandardFdroidDebugUnitTest
... (Debug only)
```

The old `testStandardGoogleplayRelease` task **no longer exists** in AGP 9.0. This is an AGP 9.0 change - unit tests now only run against debug builds, which makes sense because:
1. Debug builds don't require signing
2. Better debug symbols
3. Release builds have R8 optimizations that can complicate testing

So the CI change is technically correct and necessary. However, it probably should have been part of the AGP 9.0 migration commit (14937b8cd) rather than this PR. But since it's needed anyway, it can stay here.

✅ CI change approved.

PR #3623, opened 2026-01-29
Feature/second search app

loosestrife:
second app users can optionally install to get their search the Bible context menu extensions


tuomas2:
Thanks for the PR! The idea of a separate companion app for PROCESS_TEXT is actually a good approach — we previously had PROCESS_TEXT support built into AndBible directly but removed it because it caused more complaints than it provided value. A separate optional app is a nice solution.

However, we wouldn't want to include the companion app module in the AndBible repository itself — that should be published as a separate project that you maintain. What we *can* accept are the small changes on the AndBible side that enable your app to work:

1. **Manifest change**: Making the `Search` activity exported with the `net.bible.android.action.SEARCH` intent filter — this is fine.

2. **Search.kt change**: Handling the `search_string` extra — this is fine in principle, but there's a bug: your code is inserted *before* `loadSelectedTranslations()` is called, so `onSearch()` would run with an empty `selectedTranslations` list. The intent handling should be moved to the end of `onCreate()` (after all initialization is complete).

3. **build.gradle.kts and settings.gradle.kts changes**: These are only needed for the companion app module and should be removed from this PR.

So in summary: could you update this PR to only include the `AndroidManifest.xml` and `Search.kt` changes (without the `andbiblecontextdataextensions` module), and move the search_string handling to the end of `onCreate()`? That would be a very small, clean change we can merge.

PR #3628, opened 2026-02-02
Added check to always use an available bible to open cross references…

### LINKED ISSUE #3609: If links window has had epub file open, it tries to open also bible verses in epub
User is reporting 

> After opening a Bible window:
> 
> 1) Select a cross-reference in Bible text
> 2) Select the link in the cross-reference pop-up window 
> 3) A second pane opens trying to find that reference in an epub book instead if a Bible text window. A message is displayed: "Ps 77:6 was not found in document ...epub"

Image

https://support.andbible.org/scp/tickets.php?id=3075

Tritonfc:
Fixes #3609

### DIFF
--- a/app/src/main/java/net/bible/android/control/link/LinkControl.kt
+++ b/app/src/main/java/net/bible/android/control/link/LinkControl.kt
@@ -387,7 +387,7 @@ class LinkControl @Inject constructor(
         if (windowMode == WindowMode.WINDOW_MODE_NEW) {
             windowControl.addNewWindow(document?: defaultDocument, key)
         } else if (checkIfOpenLinksInDedicatedWindow() && !forceOpenHere) {
-            windowControl.showLink(document, key)
+            windowControl.showLink(document ?: defaultDocument, key)
         } else { // old style - open links in current window
             currentPageManager.setCurrentDocumentAndKey(document ?: defaultDocument, key)
         }
--- a/app/src/main/java/net/bible/android/control/page/window/WindowControl.kt
+++ b/app/src/main/java/net/bible/android/control/page/window/WindowControl.kt
@@ -102,13 +102,7 @@ open class WindowControl @Inject constructor() {
             linksWindow.windowState = WindowState.VISIBLE
         }
 
-        // Ensure we have a proper Bible document for verse keys to prevent
-        // trying to look up Bible verses in non-Bible documents (like EPUBs)
-        val actualDocument = document ?: if (key is VerseKey<*>) {
-            activeWindow.pageManager.currentBible.currentDocument ?: firstBibleDoc
-        } else null
-
-        linksWindow.pageManager.setCurrentDocumentAndKey(actualDocument, key)
+        linksWindow.pageManager.setCurrentDocumentAndKey(document, key)
 
         if (!linksWindowWasVisible) {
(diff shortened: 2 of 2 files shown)


tuomas2:
Thanks for the fix! The root cause analysis is correct — when `document` is null and the key is a `VerseKey`, the links window falls back to whatever document it had before (e.g. an EPUB).

However, I think there's a simpler and more consistent fix. The actual oversight is in `LinkControl.showLink()` (line 390), where the `?: defaultDocument` fallback is missing — compare with lines 388 and 392 which already have it:

```kotlin
fun showLink(document: Book?, key: Key, forceOpenHere: Boolean = false) {
    val currentPageManager = currentPageManager
    val defaultDocument = currentPageManager.currentBible.currentDocument!!
    if (windowMode == WindowMode.WINDOW_MODE_NEW) {
        windowControl.addNewWindow(document ?: defaultDocument, key)           // ✓ has fallback
    } else if (checkIfOpenLinksInDedicatedWindow() && !forceOpenHere) {
        windowControl.showLink(document, key)                                  // ✗ missing fallback
    } else {
        currentPageManager.setCurrentDocumentAndKey(document ?: defaultDocument, key)  // ✓ has fallback
    }
}
```

So the fix could simply be:
```kotlin
windowControl.showLink(document ?: defaultDocument, key)
```

This would be more consistent with the other two code paths in the same method, and it also avoids a subtle issue in the current PR: when the links window wasn't previously visible, `activeWindow` gets reassigned to the links window (line 100) *before* your new code runs, so `activeWindow.pageManager.currentBible.currentDocument` reads from the links window's Bible page rather than the source window's.

Tritonfc:
> Thanks for the fix! The root cause analysis is correct — when `document` is null and the key is a `VerseKey`, the links window falls back to whatever document it had before (e.g. an EPUB).
> 
> However, I think there's a simpler and more consistent fix. The actual oversight is in `LinkControl.showLink()` (line 390), where the `?: defaultDocument` fallback is missing — compare with lines 388 and 392 which already have it:
> 
> ```kotlin
> fun showLink(document: Book?, key: Key, forceOpenHere: Boolean = false) {
>     val currentPageManager = currentPageManager
>     val defaultDocument = currentPageManager.currentBible.currentDocument!!
>     if (windowMode == WindowMode.WINDOW_MODE_NEW) {
>         windowControl.addNewWindow(document ?: defaultDocument, key)           // ✓ has fallback
>     } else if (checkIfOpenLinksInDedicatedWindow() && !forceOpenHere) {
>         windowControl.showLink(document, key)                                  // ✗ missing fallback
>     } else {
>         currentPageManager.setCurrentDocumentAndKey(document ?: defaultDocument, key)  // ✓ has fallback
>     }
> }
> ```
> 
> So the fix could simply be:
> 
> ```kotlin
> windowControl.showLink(document ?: defaultDocument, key)
> ```
> 
> This would be more consistent with the other two code paths in the same method, and it also avoids a subtle issue in the current PR: when the links window wasn't previously visible, `activeWindow` gets reassigned to the links window (line 100) _before_ your new code runs, so `activeWindow.pageManager.currentBible.currentDocument` reads from the links window's Bible page rather than the source window's.

Hmm yeah. Just noticed there are multiple places where the links flow passes through.

PR #3658, opened 2026-02-25
Feature/#3631 show end position

### LINKED ISSUE #3631: Show end position in browsing history
## Feature Request

Show both the start and end position in the browsing history, not just the starting verse.

### Use Case

When browsing the Bible during a church service or Bible study, the user navigates through multiple passages. Currently, the history only shows where each session started. The user would like to also see where they ended up (e.g., "Matt 5:1–48") so they can:

1. Recall which passages they read through
2. Easily find and return to passages looked up during a Bible study session

Currently, to save a position in history, one must close and reopen the app before navigating to a new passage — which is cumbersome.

### Device Info
- App version: 5.0.902
- Device: Samsung SM-S901B (Android 16, SDK 36)

OSTicket: https://support.andbible.org/scp/tickets.php?id=3134

Tritonfc:
closes #3631

### DIFF
--- a/app/schemas/net.bible.android.database.WorkspaceDatabase/14.json
+++ b/app/schemas/net.bible.android.database.WorkspaceDatabase/14.json
@@ -1,0 +1,1028 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 14,
+    "identityHash": "f2a6985eccbecbb680a60afa7f24a3c8",
+    "entities": [
+      {
+        "tableName": "Workspace",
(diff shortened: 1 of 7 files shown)


Tritonfc:
I wonder how this change affects other users who may or may not care much about the feature. Currently this always stores the start and end positions. Maybe they will like it?

tuomas2:
## Review

### Positive aspects

- **`originalKey` fix is correct and well-explained.** Moving `originalKey = null` after `localSetCurrentDocument()` is a genuine bug fix — the previous code set null before the call that immediately overwrote it with Genesis 1:1. The comment explains the reasoning well.
- **DB migration is properly versioned** (8→9) with a null default, ensuring backward compatibility.
- **`formatRangeDescription` handles edge cases cleanly**: same-chapter, cross-chapter, cross-book, and end ≤ start (falls back to start-only). Error handling with try/catch is appropriate since JSword can throw on verse lookups.
- **`equals()` correctly excludes `endKey`** (compares only `document.initials + key`), which is exactly what the `else if` stack-update branch relies on.

---

### Issues

**1. `createdAt` gets a fresh timestamp when `endKey` is updated (medium)**

In `add()`, when the same document+start verse is revisited with a new `endKey`, the existing item is popped and the new item is pushed. The new item's `createdAt` defaults to `Date(System.currentTimeMillis())`, so the history entry's timestamp updates every time the user scrolls further through the same passage. The original timestamp should be preserved:

```kotlin
} else if (item is KeyHistoryItem && stack.peek() is KeyHistoryItem) {
    val existing = stack.peek() as KeyHistoryItem
    val newEndKey = item.endKey
    if (newEndKey != null && newEndKey != existing.endKey) {
        stack.pop()
        stack.push(item)  // item.createdAt is always "now" — should be existing.createdAt
    }
}
```

**2. No test coverage (medium)**

`formatRangeDescription()` and the `endKey`-update branch in `add()` are new non-trivial logic paths with no unit tests. At minimum, unit tests for the range description formatting (same chapter, cross-chapter, cross-book, end before start) would be valuable.

**3. Inaccurate comment on `endAnchorOrdinal` field (minor)**

```kotlin
// End position is set when the user navigates away from this history entry
var endAnchorOrdinal: OrdinalRange? = null
```

This field is not dynamically set when the user navigates away — it is set during DB restoration. Worth correcting.

PR #3669, opened 2026-02-27
Add search history dropdowns to search screens

### LINKED ISSUE #831: Search history
Old item lookup similar way as language selector in download documents view.

Also:
- search bookmarks
- search studypads (labels)

ghost:
## What changed
- Replaced search inputs in Bible Search and EPUB Search with `AutoCompleteTextView`.
- Added persisted search history storage (`SearchHistoryStore`) for both Bible and EPUB search.
- Added history dropdown behavior on focus/click for quick reuse of previous queries.
- Added unit tests for history update/deserialize logic.

## Benefits
- Users can quickly reuse previous search queries from a dropdown list.
- Query history is persisted between app sessions.
- Keeps issue #831 intent (old item lookup) aligned with language-selector style interaction.

## Possible side effects
- Search fields now show dropdown suggestions when focused/clicked if history exists.
- Very old history entries are truncated (keeps latest 20 entries).

## Validation
- `./gradlew :app:testStandardGithubDebugUnitTest --tests net.bible.android.view.activity.search.SearchHistoryStoreTest -x jsBuild --no-daemon`
- `./gradlew :app:assembleStandardGithubDebug -x jsBuild --no-daemon`

## Screenshots
- No structural visual UI changes; behavior update is dropdown suggestions in search fields.

Fixes #831 (Search history)

### DIFF
--- a/app/src/main/java/net/bible/android/view/activity/search/EpubSearch.kt
+++ b/app/src/main/java/net/bible/android/view/activity/search/EpubSearch.kt
@@ -30,7 +30,6 @@ import android.view.KeyEvent
 import android.view.Menu
 import android.view.MenuItem
 import android.view.inputmethod.EditorInfo
-import android.widget.ArrayAdapter
 import android.widget.TextView
 
 import net.bible.android.activity.R
@@ -51,7 +50,6 @@ import javax.inject.Inject
 class EpubSearch : CustomTitlebarActivityBase(R.menu.search_actionbar_menu) {
 
     private lateinit var binding: EpubSearchBinding
-    private lateinit var searchHistoryAdapter: ArrayAdapter<String>
     override val integrateWithHistoryManager: Boolean = true
 
     @Inject lateinit var pageControl: PageControl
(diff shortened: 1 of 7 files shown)


tuomas2 (commented):
Thanks for the contribution! The feature works well and `SearchHistoryStore` is cleanly designed. A few things to address:

## Code duplication

`setupSearchHistoryDropdown` is copy-pasted almost identically between `Search.kt` and `EpubSearch.kt`. Consider extracting it as an extension function, e.g.:

```kotlin
private fun AutoCompleteTextView.setupSearchHistory(
    history: List<String>,
    onAddQuery: (String) -> Unit
) { ... }
```

## `setOnClickListener` may be redundant

`AutoCompleteTextView` already shows the dropdown on click/touch internally via `onTouchEvent`. With `completionThreshold="0"` the dropdown opens on focus anyway. The explicit `setOnClickListener` is likely unnecessary and could interfere with the view's native behaviour on some Android versions.

## Tests

**First test doesn't clearly demonstrate "prepend" behaviour:**

```kotlin
existingHistory = listOf("john", "romans 8"),
rawQuery = "  john  "
// result is the same list — "john" was already first
```

Starting with `listOf("romans 8", "john")` would make it clear that the existing entry is promoted to the front.

**Missing edge case — separator character in query:**

```kotlin
// \u001F in a query should be sanitised before storage
SearchHistoryStore.updateHistory(emptyList(), "query\u001Fwith\u001Fseparator")
// Expected: ["query with separator"]
```

The sanitisation code is already there (`replace(ENTRY_SEPARATOR, " ")`), it just isn't tested.

---

Otherwise the design is solid — good separator choice (`\u001F`), defensive injection prevention, clean use of `internal` for testability. Nice work!

PR #3682, opened 2026-03-05
tuomas/tracking

tuomas2:

**Screenshots**


app/bibleview-js/src/components/memorize/WordScramble.vue
meichthys, line 108: I think here we need to check if the scrambled word game was completed perfectly. Just completing the scrambled word game once should not mean the user has memorized the verse or a passage. I'm not sure if we currently track when the scrambled word game has a failed attempt, but we probably should - otherwise, verses will be getting marked as memorized way to easily.

It seems as though the app knows when the user makes a mistake since the selected word, if selected incorrectly, is highlighted in red: 



 Maybe we can count these "incorrect" events and show the number of incorrect selections next to the word scramble game title (like a red notification dot with the incorrect number in the dot). We can then use that number to determine if the verse was memorized correctly: 



I also think that by clicking the peak button in the word scramble game that should also disqualify the verse from being memorized (to keep it simple, the peak button could just count as an incorrect selection).

PR #3794, opened 2026-05-04
Track multiple chapter reads

tuomas2:

**Screenshots**


agrogers:
The color scale was harder than I thought it would be to get right... and I wasnt real happy with my approach. It probably needs to be thought through more than I did. 

My ideas:
* Most books will be read less than 100%
* A few books might be read many times (eg Jude) - say  10 times - 1000%
* This messes with the colouring because most books will get squashed into a very small band 10% of the scale and Jude takes up the rest. That seems wrong.
* My idea (not implemented) was that 100% is always in the middle or above the middle of the scale. So if all books are below 100%, 100% will be at the scale maximum
* At 200%+, 100% will still be fixed to the middle and the rest of the scale fits into the remaining half
* The colors remain the same - a fixed range of colours from 0 to 100% and then a maximum color which represents whatever the maximum currently is and then a colour ramp between them.

I think the same sort of thing happens with chapters. The implementation is supposed to fix the maximum reads at 10. Perhaps the same could be done - 10 never falls below the middle - that would still give deliniation for most chapter reads which are below 10.

